PR #5, opened 2016-02-02
Basic intrinsics

mrigger:
This changes provides some Java substitutions for native calls to malloc, sqrt, and free.


projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVMOptimizations.java
```diff
@@ -54,4 +54,9 @@ private LLVMOptimizations() {
      */
     public static final boolean BRANCH_INJECTION_CONDITIONAL_BRANCH = true;
 
+    /**
+     * Substitute common C library functions by Java implementations. Note that some functions such
+     * as exit or abort are substituted in each case, to allow the VM to terminate gracefully.
+     */
+    public static final boolean INTRINSIFY_C_LIBRARY_FUNCTIONS = true;
```
chrisseaton: We'll probably want to pull this out into a simple configuration system soon.

chrisseaton:
You could have a single class `MemoryIntrinsics` with `malloc` and `free` as nested classes, as they're so simple, and then a `MathInstrinsic` class for `sqrt` and the others when you do them.

mrigger:
I plan to refactor intrinsics soon anyway. Instead of getting the arguments array by themselves, they should receive typed arguments in their execute method. Also, it should be configurable if an intrinsic should be a call target or just a node in order to make decisions about inlining upfront. Then, the nodes should inherit from the node class with respect to their return type. For experimentation, it would be nice to leave them for now like this.

PR #6, opened 2016-02-03
set optimization options with system properties

mrigger:
This change lets a user set optimization flags over Java system properties. Whether an optimization is enabled is represented by a Configuration Object so that later on the Java system properties can be transparently replaced by a per Sulong instance option system.

I also started refactoring the option system, with more fixes to come.


chrisseaton:
Does this implementation allow for options to be final in the PE? Or is that never needed?

mrigger:
That is never needed. The optimization options are just used to decide which nodes should be instantiated during AST construction time or during linking.

PR #9, opened 2016-02-04
value profiling for float, double, and remaining integer types

mrigger:
Previously, value profiling for memory loads has only been performed for I32 and I64 types. This changes adds value profiling for float, double, and the remaining integer types. There are still other loads which are not profiled because it is unclear if there is a benefit (e.g., for pointers).

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMLoadDoubleNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMLoadDoubleNode.java
@@ -81,7 +81,7 @@ public abstract class LLVMLoadDoubleNode extends LLVMDoubleNode {
         @Override
         public double executeDouble(VirtualFrame frame) {
             double value = LLVMMemory.getDouble(addressNode.executePointee(frame));
-            if (value == profiledValue) {
+            if (Double.doubleToRawLongBits(value) == Double.doubleToRawLongBits(profiledValue)) {
                 return profiledValue;
             } else {
                 CompilerDirectives.transferToInterpreterAndInvalidate();
--- a/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMLoadDoubleNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMLoadDoubleNode.java
@@ -29,67 +29,65 @@
  */
 package com.oracle.truffle.llvm.nodes.memory.load;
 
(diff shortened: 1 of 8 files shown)


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMLoadDoubleNode.java
```diff
@@ -0,0 +1,94 @@
...
+
+        public LLVMLoadValueProfiledDoubleNode(LLVMAddressNode addressNode, double profiledValue) {
+            this.addressNode = addressNode;
+            this.profiledValue = profiledValue;
+        }
+
+        @Override
+        public double executeDouble(VirtualFrame frame) {
+            double value = LLVMMemory.getDouble(addressNode.executePointee(frame));
+            if (value == profiledValue) {
```
chrisseaton: Using `==` here might cause you problems. For example `-0.0 == 0.0`, but it's not the case for every operator that `f(-0.0) == f(0.0)`. You probably want to compare the bits instead, using `Double.toLongBits`.

However, Truffle already has classes for value profiling (which gets this right). Maybe use those?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMLoadFloatNode.java
```diff
@@ -0,0 +1,94 @@
...
+
+        public LLVMLoadValueProfiledFloatNode(LLVMAddressNode addressNode, float profiledValue) {
+            this.addressNode = addressNode;
+            this.profiledValue = profiledValue;
+        }
+
+        @Override
+        public float executeFloat(VirtualFrame frame) {
+            float value = LLVMMemory.getFloat(addressNode.executePointee(frame));
+            if (value == profiledValue) {
```
chrisseaton: `==` with `float` has the same problem as `double`

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMLoadDoubleNode.java
mrigger: Good point! I would suggest that I first push the fix for this, and then I will finish the travis configuration. When the gate can run the test cases I will refactor the load nodes to use the Truffle value profiling classes.

lukasstadler: if you do this yourself, you should use doubleToRawLongBits - it doesn't normalize NaNs.

PR #12, opened 2016-02-09
Simplify intrinsic implementation.

mrigger:
The goal of this change is to simplify the intrinsic implementation, for example, when the C malloc should be substituted by a Java node that implements the malloc operation in Java. Instead of manually having to read and cast arguments in the intrinsified function node, the correct argument read nodes should be inferred from the node. Additionally, the return type of the intrinsic nodes now reflect the real return types. In the long term, we have to think how to let other languages implement intrinsics for Sulong. For now, this change helps to avoid having to re-implement the intrinsic nodes in the future.


mrigger:
Since the Travis gate does not work yet, I verified the changes locally by running all Sulong test cases.

PR #13, opened 2016-02-10
intrinsify some of the C math.h functions

mrigger:
This change intrinsifies some of the C maths functions to improve performance on benchmarks such as whetstone. I am not sure whether the semantics of the Java Math methods are exactly the same as the C functions. The calls to Math delegate to StrictMath. The methods are then implemented on the native side. The StrictMath documentation states:

> The Java math library is defined with respect to fdlibm version 5.3. Where fdlibm provides more than one definition for a function (such as acos), use the "IEEE 754 core function" version (residing in a file whose name begins with the letter e). The methods which require fdlibm semantics are sin, cos, tan, asin, acos, atan, exp, log, log10, cbrt, atan2, pow, sinh, cosh, tanh, hypot, expm1, and log1p. 

For now, all test cases still pass. I would suggest to let the functions intrinsified as long as we do not run into problems. If we run into problems, we can easily determine whether they stem from the intrinsics by turning them off.

### DIFF
--- a/projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/intrinsics/c/LLVMCMathsIntrinsics.java
+++ b/projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/intrinsics/c/LLVMCMathsIntrinsics.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are
+ * permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list of
+ * conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this list of
+ * conditions and the following disclaimer in the documentation and/or other materials provided
+ * with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may be used to
+ * endorse or promote products derived from this software without specific prior written
+ * permission.
+ *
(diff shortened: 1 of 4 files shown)


projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/intrinsics/c/LLVMCMathsIntrinsics.java
```diff
@@ -0,0 +1,147 @@
...
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.llvm.LLVMIntrinsic.LLVMDoubleIntrinsic;
+import com.oracle.truffle.llvm.nodes.base.floating.LLVMDoubleNode;
+
+/**
+ * C functions from math.h
+ */
+public abstract class LLVMCMathsIntrinsics {
+
+    // acos
```
chrisseaton: Do we need these comments?

mrigger: I think it is easier to read through the file like this. One can then also quickly see that the functions are alphabetically ordered. Of course, they are redundant. Should I delete them?

chrisseaton: I think they're redundant. The class is already named `LLVMACos` so the comment provides no extra information so it's only noise. It's just a suggestion though, feel free to ignore as long as you've thought about it.

mrigger: I removed the comments. Thanks for the feedback!

PR #25, opened 2016-02-12
use Truffle API for value profiling

mrigger:
Addresses #9 by using Truffle's value profiling API where possible.


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/load/LLVMDirectLoadNode.java
```diff
@@ -43,11 +43,11 @@
 import com.oracle.truffle.llvm.types.memory.LLVMHeap;
 import com.oracle.truffle.llvm.types.memory.LLVMMemory;
 
-public abstract class LLVMLoadDirectNode {
+public abstract class LLVMDirectLoadNode {
```
chrisseaton: These kind of stylistic changes are fine, but try to separate them from semantic changes in commits.

mrigger: Good point!

PR #30, opened 2016-02-16
constructor support and passing of external function pointers to nati…

mrigger:
This change adds support C++ constructors by implementing LLVM [constructor calls](http://llvm.org/docs/LangRef.html#the-llvm-global-ctors-global-variable). Also, this change made it possible to have external function pointers in the program that can be passed to other native functions.


mrigger:
All tests successfully passed.

PR #61, opened 2016-03-01
@SuppressWarnings for deprecated Truffle method

mrigger:


chrisseaton:
I think this method has a default implementation now. You should be able to just remove it if you have a recent Truffle.

mrigger:
Ok, thanks!

PR #74, opened 2016-03-08
call memcpy over native handle as PE fail workaround

mrigger:
As analyzed by @christianwimmer, the PE currently fails for calls such as Unsafe.copyMemory used to implement memcpy due to missing LoopExits that should be inserted by the graph decoder on exception edges. This change calls the native memcpy until the issue is fixed in the partial evaluator.

### DIFF
--- a/projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMHeap.java
+++ b/projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMHeap.java
@@ -29,10 +29,15 @@
  */
 package com.oracle.truffle.llvm.types.memory;
 
+import com.oracle.nfi.NativeFunctionInterfaceRuntime;
+import com.oracle.nfi.api.NativeFunctionHandle;
+import com.oracle.nfi.api.NativeFunctionInterface;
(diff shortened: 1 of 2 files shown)


projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMHeap.java
```diff
@@ -29,11 +29,16 @@
...
 
+import com.oracle.nfi.NativeFunctionInterfaceRuntime;
+import com.oracle.nfi.api.NativeFunctionHandle;
+import com.oracle.nfi.api.NativeFunctionInterface;
 import com.oracle.truffle.llvm.types.LLVMAddress;
 import com.oracle.truffle.llvm.types.LLVMFunction;
 
 public class LLVMHeap extends LLVMMemory {
 
+    private static NativeFunctionHandle memCopyHandle;
```
rschatz: Shouldn't this be final? Then the compiler should be able to intrinsify the call to the function...

mrigger: Yes, thanks for spotting that!

PR #77, opened 2016-03-09
move phi write nodes into terminator instruction nodes

mrigger:
Conditional phi writes occur in conditional branches of a basic blick, when a successor of the basic block refers to it in a phi function. Previously, the check on the condition of the conditional branch was executed twice, namely once for the conditional phi write nodes and once for the conditional branch. @rschatz found that the redundant checks remain in the compiled code. This change moves the conditional phi write nodes together with the conditional branches, so that the condition is only checked once.


rschatz:
lgtm

PR #78, opened 2016-03-09
call memcpy over native handle as PE fail workaround

mrigger:
As analyzed by @christianwimmer, the PE currently fails for calls such as Unsafe.copyMemory used to implement memcpy due to missing LoopExits that should be inserted by the graph decoder on exception edges. This change calls the native memcpy until the issue is fixed in the partial evaluator.


rschatz:
lgtm

PR #83, opened 2016-03-10
perform some opt optimizations in the benchmarks and disable the C intrinsifications

mrigger:
I'm still trying to figure out which LLVM passes improve the performance for Sulong. So far, it seems that only the loop invariant code modition pass (`-licm`) shows improvements for the benchmarks I'm testing. I still added some other passes that could help, and did not regress performance at least.

I also disable the intrinsifications of C functions, since it seems that substituting them with Java equivalents makes Sulong slower.


lukasstadler:
I think that performance shouldn't be the only criteria - it's also important that the optimization passes preserve enough debugging information.

mrigger:
This change only affects the benchmark runner invoked by `mx suoptbench`.

PR #89, opened 2016-03-11
split travis test case execution into two parts

mrigger:
Previously, the Travis build step for executing the test cases for Sulong needed over 40 minutes and approached the maximum limit of 50 minutes that is allowed for a Travis build step. This change splits the execution of the test cases into two separate build steps. Currently, there is one for the GCC test cases (test step 2), and one for the remaining test cases (test step 1). Since the steps can run in parallel, the wall time for the Travis gate for the PR improved from about 41 minutes to 31 minutes. Since the native and Java code has to be compiled twice, the total time increased from 51 minutes to 1 hour and 8 minutes.


eregon:
Looks good. It would be nice to have the build and the test phases separate to easily monitor their respective duration.

PR #108, opened 2016-03-21
enable Travis Mac builds

mrigger:


chrisseaton:
What's going on with this issue?

mrigger:
It seems that the status is stuck at the Travis check with "Waiting for status to be reported", while the Travis build is not started.

PR #112, opened 2016-03-24
Pe workaround fix

mrigger:
#78 introduced a workaround for PE fails for memcpy. Since the PE fails were fixed in https://github.com/graalvm/graal-core/pull/160 this change removes the workaround again.

### DIFF
--- a/projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMHeap.java
+++ b/projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMHeap.java
@@ -29,15 +29,10 @@
  */
 package com.oracle.truffle.llvm.types.memory;
 
-import com.oracle.nfi.NativeFunctionInterfaceRuntime;
-import com.oracle.nfi.api.NativeFunctionHandle;
-import com.oracle.nfi.api.NativeFunctionInterface;
 import com.oracle.truffle.llvm.types.LLVMAddress;
 import com.oracle.truffle.llvm.types.LLVMFunction;
 
 public class LLVMHeap extends LLVMMemory {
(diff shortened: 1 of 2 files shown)


projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMHeap.java
```diff
@@ -66,22 +61,10 @@ public static void freeMemory(LLVMAddress addr) {
         UNSAFE.freeMemory(extractAddrNullPointerAllowed(addr));
     }
 
-    // setters
-
     public static void memCopy(LLVMAddress target, LLVMAddress source, long length) {
         long targetAddress = extractAddrNullPointerAllowed(target);
         long sourceAddress = extractAddrNullPointerAllowed(source);
```
woess: why are null pointers allowed here?

mrigger: If I remember correctly, at least one GCC test case fails if we do not allow null pointers at this location. Both the C standard library and LLVM intrinsic do not state whether null pointers are allowed, so behavior is undefined like in the other cases that do not allow null pointers. We still have the assertions in the other cases, since it is convenient to detect NULL pointers without crashing the JVM.

woess: well, maybe if length == 0, otherwise this is bound to crash the vm.

mrigger: I will add an additional assert for that.

PR #113, opened 2016-03-26
Only link libraries if -lgmp or -lgfortran are specified

chrisseaton:
This is a step towards solving #107

We currently always try to link `libgfortran` and `libgmp`. We can't just keep adding more and more libraries for every language we want to support long term, so let's only load `libc` and `libstdc++` by default, and let the user specify `-lgfortran` or `-lgmp` if they want those libraries.

Another immediate benefit of this is that `libc` and `libstdc++` are available by default on Mac, so that solves our problem of not being able to find those libraries, and running the example C program now just works for me.

### DIFF
--- a/mx.sulong/mx_sulong.py
+++ b/mx.sulong/mx_sulong.py
@@ -364,8 +364,15 @@ def runDebugLLVM(args=None):
 
 def runLLVM(args=None):
     """uses Sulong to execute a LLVM IR file"""
-    vmArgs, sulongArgs = truffle_extract_VM_args(args)
-    return mx.run_java(getCommonOptions() + vmArgs + ['-XX:-UseJVMCIClassLoader', '-cp', mx.classpath(['com.oracle.truffle.llvm']), "com.oracle.truffle.llvm.LLVM"] + sulongArgs, jdk=mx.get_jdk(tag='jvmci'))
+    vmArgs, sulongArgsWithLibs = truffle_extract_VM_args(args)
+    sulongArgs = []
+    libNames = []
+    for arg in sulongArgsWithLibs:
+        if arg.startswith('-l'):
+            libNames.append(arg)
+        else:
+            sulongArgs.append(arg)
+    return mx.run_java(getCommonOptions(libNames) + vmArgs + ['-XX:-UseJVMCIClassLoader', '-cp', mx.classpath(['com.oracle.truffle.llvm']), "com.oracle.truffle.llvm.LLVM"] + sulongArgs, jdk=mx.get_jdk(tag='jvmci'))
 
 def runTests(args=None):
     """runs all the test cases"""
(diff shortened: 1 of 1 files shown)


chrisseaton:
@bjfish can you try this?

chrisseaton:
I added instructions to the `README.md` that tell you how to get the other libraries, but of course you don't need them now unless you are actually going to use them.

bjfish:
@chrisseaton Yes, this works for me. I was able to run `mx su-run test.ll`. Thanks!

mx.sulong/mx_sulong.py
```diff
@@ -419,13 +426,30 @@ def runTypeTestCases(args=None):
     vmArgs, _ = truffle_extract_VM_args(args)
     return unittest(getCommonUnitTestOptions() + vmArgs + ['com.oracle.truffle.llvm.types.floating.test'])
 
-def getCommonOptions():
-    return ['-Dgraal.TruffleCompilationExceptionsArePrinted=true', '-Dgraal.ExitVMOnException=true', '-Dsulong.IntrinsifyCFunctions=false', getSearchPathOption()]
+default_lib_args = ['-lgmp', '-lgfortran']
+
+def getCommonOptions(lib_args=default_lib_args):
```
mrigger: I think that the [suOptBench](https://github.com/chrisseaton/sulong/blob/library-options/mx.sulong/mx_sulong.py#L542) benchmark runner should link all the libraries by default to make the automatic execution of the benchmarks easier. I can also address this in a subsequent PR.

README.md
```diff
@@ -25,6 +25,18 @@ Make sure you have GCC-4.6, G++-4.6, and GFortran-4.6 installed. For
 a full list of external dependencies on Ubuntu you can look at our
 [Travis configuration](https://github.com/graalvm/sulong/blob/master/.travis.yml).
 
+On the Mac you can use Homebrew:
+
+    brew install gcc
```
mrigger: Dragonegg supports GCC versions [up to 4.8](http://dragonegg.llvm.org/). Although you get the right Fortran library with GCC 5.3, it would still be better to install a version that works with Dragonegg.

mx.sulong/mx_sulong.py
chrisseaton: I think it already does, doesn't it? It calls `getSearchPathOption()` with no argument, so it gets `default_lib_args`, which is `['-lgmp', '-lgfortran']`.

mrigger: Yes, you are right. I overlooked that.

README.md
chrisseaton: Unfortunately there are no files with `fortran` in the name at all installed by GCC 4.8 :(

mrigger: Under Ubuntu I have to install a separate package `gfortran-4.6`. Maybe the Homebrew `gcc` package just groups several GCC projects together?

mrigger: We can also address this in another PR.

chrisseaton: Ok, I've removed the `README` stuff, so it's now just the library change. I think I almost have the right set of flags for Mac and Homebrew now, so I'll keep working on that and do a separate PR.

Issue #114, opened 2016-03-27
Commit Messages and PR descriptions


mrigger:
Since the project is steadily growing, I propose a stricter approach to commit messages and PR request descriptions to increase maintainability and speed up reviewing. I have the following suggestions:

Commit messages
- Each commit (message) should contain (and describe) one logical change.

PR request:
- Each PR request should address one feature or change.
- Self-explanatory changes such as updates of the Graal version can stay unexplained.
- Other changes should at least describe why (1) the change or feature is needed, (2) how the change or feature is implemented, and optionally (3) what further implications the change has. Personally, I prefer the Github description field to describe a change in more detail. Alternatively, this could also be part of a commit message.

Both PR request titles and commit messages  (first two bullets copied from [here](https://github.com/erlang/otp/wiki/writing-good-commit-messages)):
- Write the summary line and description of what you have done in the imperative mode, that is as if you were commanding someone. Start the line with "Fix", "Add", "Change" instead of "Fixed", "Added", "Changed".
- Don't end the summary line with a period - it's a title and titles don't end with a period.
- Each title and commit message should start with a capital letter.

Links:
- https://github.com/erlang/otp/wiki/writing-good-commit-messages
- http://who-t.blogspot.co.at/2009/12/on-commit-messages.html
- http://chris.beams.io/posts/git-commit/

mrigger:
Since nobody complained about these rules, I will document them in the CONTRIBUTING file. Please reject future commits that do not comply with these rules.

PR #136, opened 2016-04-10
Commit message checks

mrigger:
We decided to use stricter commit message and PR description rules in #114. To help enforcing this, this change introduces the following checks on Git commit message:

1. A message cannot start with a lower case character.
2. A message cannot end with a period, exclamation mark, or question mark.
3. A commit message should not contain past tense words such as "implemented", "fixed", "merged", and others.

Currently, many old commit messages do not comply do these rules. Thus, currently only the latest ten commit messages are checked.

Here is an example output caused by some recent violations:

```
"update Graal version" starts with a lower case character
"Update Mac building instructions." ends with period, question mark, or exclamation mark
"execute program five times in the benchmark runner" starts with a lower case character
"unreachable node documentation" starts with a lower case character
"fix assertion in insert value node construction" starts with a lower case character
"move phi node creation into the node facade" starts with a lower case character
"use constant evaluator for insert value constant evaluation" starts with a lower case character
"move memcpy node to the facade" starts with a lower case character
"remove unused conditional phi write node" starts with a lower case character
"remove untested metadata and debug nodes" starts with a lower case character
"remove untested invoke and landingpad prototypes" starts with a lower case character
"remove unused imports" starts with a lower case character
 
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ branches:
   only:
     - master
 env:
+  - TEST_COMMAND='mx/mx su-gitlogcheck'
   - TEST_COMMAND='mx/mx pylint'
   - TEST_COMMAND='mx/mx checkstyle'
   - TEST_COMMAND='mx/mx checkoverlap'
--- a/mx.sulong/mx_sulong.py
+++ b/mx.sulong/mx_sulong.py
@@ -620,7 +620,7 @@ def standardLinkerCommands(args=None):
 import subprocess
 import re
 
-titleWithEndingPunct = re.compile('^(.*)[\.!?]$')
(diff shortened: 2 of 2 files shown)


chrisseaton:
> Currently, many old commit messages do not comply do these rules. Thus, currently only the latest ten commit messages are checked.

Maybe only check commits later than today's date? Many PRs could be much larger than 10 commits, and actually those are probably most important to check as if there's lots of them you may not bother looking at them all manually.

mrigger:
If someone already started development based on an earlier Sulong version and does not do a rebase then these commits would also not be checked at all. Maybe we can have (as a feature in the future) a whitelist of commits that violate the checks but are already part of the history?

PR #141, opened 2016-04-11
Use boxed primitives in facade

mrigger:
With this change, the node creation facade uses boxed primitives to use null as a default value. This allows composite facades to delegate the implementation of a method to another facade.


chrisseaton:
What does `null` signify? The absence of a value? Use `Optional` in that case.

mrigger:
Do you think that `Optional` would be more elegant? A `null` value only makes sense when combining two factories with `NodeFactoryFacadeComposite` where the first factory can return a `null` value. The parser should always receive a non `null` value from the facade. Having `Optional` would thus bloat the parser code.

chrisseaton:
I think `Integer` is a code smell that you probably want `Optional`. The fact that it only makes sense in some cases makes it more sensible to make it explicit I think.

Not a strong opinion though.

Issue #137, opened 2016-04-10
Create easy-to-manage package for LLVM bitcode


chrisseaton:
Languages which can load extensions, like Ruby, are often build around a model of `.so` files being on a load path and being dynamically loaded. I'd like to be able to provide something similar for Sulong.

So what I propose is a linker command that produces a `.su` file - a Sulong library file - which is a zip file with LLVM bitcode in it, and a manifest that specifies which libraries it depends on.

Then where Ruby normally allows you to load a `.so` file, we do exactly the same thing with our `.su` files.

Linking bitcode files into a `.su` and specifying which libraries to load when that library is loaded would look like this:

```
$ mx su-link test.su -lz test.ll
```

You could then directly run the `test.su` file:

```
$ mx su-run test.su
```

Or you can load your Sulong library like you would a native library:

```
$ mx su-run -ltest main.ll
```

Sulong would automatically detect the difference between a native library and a Sulong library, and in the latter case it would just load all the dependent libraries and then all the bitcode in the zip file.

Why not just link everything into a single big bitcode file for your library? Because we need somewhere to list dependent libraries. If a Ruby extension needs `zlib` then I need to say that somewhere.

I've started working on this in https://github.com/graalvm/sulong/compare/master...chrisseaton:su-files

Let me know if you think I'm heading down the wrong path.

mrigger:
I agree that an own file format makes sense. I looked through the code and it looks fine to me. Here a few comments/thoughts:
- We could also support to accept several bitcode files by `su-link `and link the bitcode files together by using [llvm-link](http://llvm.org/docs/CommandGuide/llvm-link.html).
- Maybe we should follow the convention of GCC and Clang to specify the output file with` -o`? Then we could invoke the link command with either `mx su-link -o test.su -lz test.ll` or `mx su-link -lz test.ll -o test.su`?
- I think we should keep the manifest file extensible, so that we can add flags when loading libraries.

PR #150, opened 2016-04-13
Add enough interop and infrastructure to load Ruby C extensions

chrisseaton:
Adds a file format, `.su` which is bit code files packaged up and made to look like an `.so` or `.bundle` file so that it's the same model as real native libraries and can be handled in a similar way.

Adds a linker to produce those files.

Adds very basic interop support, enough to execute a function.

There are a few hacks and todos here that I wasn't able to fix myself. I will point them out in comments.

Fixes https://github.com/graalvm/sulong/issues/137.

Enables https://github.com/jruby/jruby/pull/3800.


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
```diff
@@ -102,4 +104,16 @@ public Source getSourceFile() {
...
+    private static LLVMContext lastContext;
+
+    public static CallTarget getCallTarget(LLVMFunction function) {
+        return lastContext.registry.lookup(function);
+    }
+
+    public static LLVMStack getStaticStack() {
+        return lastContext.stack;
+    }
+
```
chrisseaton: `LLVMContext` is not accessible from `LLVMFunction` - see code there for why we need to do it.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMLanguage.java
```diff
@@ -38,12 +38,28 @@
...
+     */
+
+    static {
+        try {
+            Class.forName("com.oracle.truffle.llvm.LLVM", true, ClassLoader.getSystemClassLoader());
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
```
chrisseaton: Another dependency order hack.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/TestPolyglotEngine.java
```diff
@@ -0,0 +1,66 @@
...
+import com.oracle.truffle.api.vm.PolyglotEngine;
+import com.oracle.truffle.api.vm.PolyglotEngine.Value;
+
+public class TestPolyglotEngine {
+
+    @Test
+    public void testExecute() throws IOException {
+        final PolyglotEngine engine = PolyglotEngine.newBuilder().build();
+        try {
+            engine.eval(Source.fromFileName("projects/com.oracle.truffle.llvm.test/tests/llvmir/simple/1.ll"));
```
chrisseaton: I couldn't work out what path to use, including using the constants from `LLVMPaths`, that would let me run these tests and find the files from both Eclipse's JUnit runner and `mx unittest`.

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMFunction.java
```diff
@@ -216,4 +289,57 @@ public boolean equals(Object obj) {
...
+                final Class<?> contextClass = Class.forName("com.oracle.truffle.llvm.nodes.impl.base.LLVMContext");
+                final Method getCallTargetMethod = contextClass.getMethod("getCallTarget", LLVMFunction.class);
+                return (CallTarget) getCallTargetMethod.invoke(null, function);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+    }
+
```
chrisseaton: This is where we need that access to `LLVMContext`.

projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
```diff
@@ -104,6 +144,20 @@ public static void main(String[] args) throws Exception {
...
+            try (FileOutputStream stream = new FileOutputStream(file.getPath())) {
+                stream.write(source.getBytes(StandardCharsets.UTF_8));
+            }
+
+            return parseFile(file.getPath(), context);
+        } finally {
+            file.delete();
+        }
+    }
+
```
chrisseaton: Couldn't figure out how to run the parser on a string without writing it out to a file.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/TestPolyglotEngine.java
mrigger: You can base the path on [LLVMPaths.java#L49](https://github.com/graalvm/sulong/blob/91ee923c780982b816480dd6397570b61e8b2d3c/projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/LLVMPaths.java#L49). The option in [mx_sulong.py#L500](https://github.com/graalvm/sulong/blob/91ee923c780982b816480dd6397570b61e8b2d3c/mx.sulong/mx_sulong.py#L500) sets the right directory for mx.

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMFunction.java
```diff
@@ -216,4 +289,57 @@ public boolean equals(Object obj) {
...
+            stack = getLLVMStack();
+            callNode = Truffle.getRuntime().createIndirectCallNode();
+        }
+
+        @Override
+        public Object execute(VirtualFrame frame) {
+            assert ForeignAccess.getArguments(frame).isEmpty();
+            final LLVMFunction function = (LLVMFunction) ForeignAccess.getReceiver(frame);
+            final CallTarget callTarget = getCallTarget(function);
+            return callNode.call(frame, callTarget, new Object[]{stack.allocate()});
```
mrigger: I believe that the stack is not freed any more and will result in a memory leak. Can we implement a getter for `LLVMStack#upperBounds` (and `#lowerBounds` for consistency) so that we use the already allocated stack memory? Currently the stack is freed in `LLVMGlobalRootNode`. Is this node used anywhere? If not, then global variables are probably not yet supported.

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMFunction.java
```diff
@@ -216,4 +289,57 @@ public boolean equals(Object obj) {
         }
     }
 
+    private static class ForeignCallNode extends RootNode {
```
mrigger: I believe we should pass a [factory facade](https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/NodeFactoryFacade.java) to each `LLVMFunction`. The `LLVMFunction` can then use a newly introduced method to instantiate a `ForeignCallNode` (maybe better: `LLVMForeignCallNode`) that is implemented in `com.oracle.truffle.llvm.nodes.impl`. We could address that in a separate PR.

projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
```diff
@@ -75,9 +80,44 @@
...
+                        }, source -> {
+                            ParserResult parserResult;
+                            try {
+                                parserResult = parseString(source.getCode(), llvmContext);
+                            } catch (IOException e) {
+                                throw new UncheckedIOException(e);
+                            }
+
+                            llvmContext.getFunctionRegistry().register(parserResult.getParsedFunctions());
+                            if (parserResult.getMainFunction() != null) {
```
mrigger: Can this be `null` in the current implementation? In the visitor you create `Truffle.getRuntime().createCallTarget(RootNode.createConstantNode(null))` if there is no main function.

projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
mrigger: I think it is okay for now to use a temporary file for that.

mrigger:
I think we can address some of these issues in a separate PR, namely:
- to remove the `lastContext` field in `LLVMContext`
- a fix for the dependency order in `LLVMLanguage`
- to move the `ForeignCallNode` to the node factory (https://github.com/graalvm/sulong/pull/149 will make that easier)

To address this issues we have to do fix architectural problems.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/TestPolyglotEngine.java
chrisseaton: Fixed.

projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
chrisseaton: Fixed.

mrigger:
We also have to address static initializers and destructors in a subsequent PR.

PR #156, opened 2016-04-14
Execute constructors and destructors for Sulong libraries

mrigger:
This change factors out the static initializers and static destructors. Instead of executing them in the `LLVMGlobalRootNode` the static initializer nodes are executed when a source file is parsed, and the static destructors are executed when the `PolyglotEngine` is disposed.

Addresses #150 and https://github.com/jruby/jruby/pull/3800.


chrisseaton:
This works - I see `printf` working from a Ruby `Init_` function.

PR #157, opened 2016-04-14
Define a potential header for interop functions

chrisseaton:
Starting to sketch out what a header file for using interop functions from C could look like. Just enough to start a conversation by showing how Ruby C extensions could compile against it.

### DIFF
--- a/include/truffle.h
+++ b/include/truffle.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are
+ * permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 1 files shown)


chrisseaton:
Truffle C used to have versions of most of these method with specific return types. We might well need that but I want to keep this simple to start with, just so I can start to sketch out what Ruby needs.

chrisseaton:
See https://github.com/jruby/jruby/pull/3803. Together with this PR, this is a starting point for whoever is going to work on Sulong interop. Lots of open questions here!

eregon:
It might be interesting to have a type like "ForeignObject"/"TruffleObject", then Sulong can know whatever it does with that is via Interop, and all functions Ruby would expose would either have simple types or that (i.e. VALUE would become ForeignObject for most cases). TruffleObject is already used for similar purposes at the Java level I think.

PR #160, opened 2016-04-17
Bitcode libraries

mrigger:
This change implements the linking of bitcode libraries. mx compiles C files in `mx.sulong/libs` to bitcode and passes them with the `sulong.DynamicBitcodeLibraries` option to the Sulong runtime. The Sulong runtime parses the libraries and registers the functions, so that the main program can use them.

### DIFF
--- a/mx.sulong/libs/test.c
+++ b/mx.sulong/libs/test.c
@@ -1,0 +1,3 @@
+int test() {
+	return 3;
+}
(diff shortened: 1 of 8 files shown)


chrisseaton:
We'll also need to do some changes for my idea for allowing libraries to be listed within `.su` files. At the moment I think the list of loaded libraries is totally static.

mrigger:
Yes, the fix for the function registry from this PR should also help to implement that. 

I did not want to have the checkstyle and pylint fix in the history, so I edit the history and pushed a new PR: #161.

chrisseaton:
You don't need to delete a PR to change the history. Just make whatever local changes and then `push --force` and it will replace those already there.

mrigger:
Good to know, thanks!

PR #162, opened 2016-04-18
Global variable aliases

mrigger:
This change adds basic support for aliases to global variables and other global variable aliases. Such aliases can be produced by GCC's `__attribute__((alias("varname")))` attribute

### DIFF
--- a/projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMParserAsserts.java
+++ b/projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMParserAsserts.java
@@ -40,4 +40,10 @@ public class LLVMParserAsserts {
         return objects;
     }
 
+    public static void assertNotNull(Object object) {
+        if (object == null) {
+            throw new AssertionError();
+        }
+    }
+
 }
--- a/projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
+++ b/projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
@@ -263,7 +263,7 @@ public class LLVMVisitor implements LLVMParserRuntime {
         List<GlobalVariable> staticVars = new ArrayList<>();
         functionToLabelMapping = new HashMap<>();
         setTargetInfo(objects);
-        allocateGlobals(this, objects);
+        allocateGlobalsAndAliases(this, objects);
         this.nativeLookup = new NativeLookup(facade);
         for (EObject object : objects) {
             if (object instanceof FunctionDef) {
@@ -287,7 +287,7 @@ public class LLVMVisitor implements LLVMParserRuntime {
             } else if (object instanceof TypeDef) {
                 // do nothing
             } else if (object instanceof Alias) {
-                // do nothing, visit later when alias is referenced
+                // do nothing
             } else if (object instanceof InlineAsm) {
                 LLVMLogger.info("ignoring module level inline assembler!");
             } else {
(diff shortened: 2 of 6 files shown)


projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMParserAsserts.java
```diff
@@ -40,4 +40,10 @@
         return objects;
     }
 
+    public static void assertNotNull(Object object) {
+        if (object == null) {
+            throw new AssertionError();
+        }
+    }
```
woess: you could use `Objects.requireNonNull` instead.

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
```diff
@@ -1006,6 +1021,15 @@ private LLVMExpressionNode visitAliasConstant(Alias ref) {
...
             LLVMFunctionDescriptor function = createLLVMFunctionFromHeader((FunctionHeader) aliaseeRef);
             return factoryFacade.createLiteral(function, LLVMBaseType.FUNCTION_ADDRESS);
+        } else if (aliaseeRef instanceof GlobalVariable) {
+            GlobalVariable originalVar = (GlobalVariable) aliaseeRef;
+            Object globalVar = globalVars.get(originalVar);
+            LLVMParserAsserts.assertNotNull(globalVar);
+            return factoryFacade.createLiteral(globalVar, LLVMBaseType.ADDRESS);
+        } else if (aliaseeRef instanceof Alias) {
+            Object globalVar = aliases.get(ref);
+            LLVMParserAsserts.assertNotNull(globalVar);
```
woess: you might want to assert that globalVar is not again an Alias.

projects/com.oracle.truffle.llvm.test/tests/c/alias/globalVar4.c
```diff
@@ -0,0 +1,17 @@
+#include <stdlib.h>
+
+int original = 1324342;
+extern int alias __attribute__((alias("original")));
+extern int alias2 __attribute__((alias("alias")));
+extern int alias3 __attribute__((alias("alias2")));
```
woess: does it also work in reverse order?

mrigger: Yes, I did not implement that since I did not need it for my use case. I added support for it in #165.

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMParserAsserts.java
mrigger: I would like to keep all asserts in this class. This would allow us, e.g., to later replace the `AssertionError` with an `assert`. I also think that it is more readable to use `LLVMParserAsserts`.

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
mrigger: I am not sure if that would be correct. The representation of an allocation is independent from the parser implementation, and implemented as class implementing the interface `NodeFactoryFacade`. It would not be impossible, that an implementation just returns global objects and aliases and resolves them during runtime.

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMParserAsserts.java
chrisseaton: But why not just `assert object != null` right now?

mrigger: I think having a central method is more flexible. As another example, we might want to selectively treat classes of errors (e.g., errors in the parser as in `LLVMParserAsserts`) differently. We already have an option `sulong.PerformanceWarningsAreFatal` where we either throw an exception or continue execution normally. We could follow a similar approach by using `LLVMParserAsserts` where we could control with an option whether we want to continue execution or throw an error.

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
woess: ok

PR #168, opened 2016-04-19
Enhance git check with an --create-ignore-file option

mrigger:
#136 introduced a basic git log message check that performs checks starting from a hard-coded date, since some previously committed files would violate the syntax checks. If someone would now commit changes with a time stamp before that date, they would not be automatically checked.

This change implements a list of exceptions of already committed files, so that we can check all log messages. With `mx su-gitlogcheck --create-ignore-file` we can create a file `gitlogfile.cfg` with the hashes of the commits that currently violate the git log checks. This file is then used when using `mx su-gitlogcheck` without options, so that these commits do not result in an error.


woess:
can't you just check all new commits instead?

mrigger:
I guess basing the check on `git log master..FETCH_HEAD` would have been simpler to implement. I was not aware of `FETCH_HEAD` before. Checking all commits has the advantage that the approach also works locally, and not just on the gate. Should we still merge it?

woess:
I was thinking only about travis; you're right, we want something that works locally, too. Therefore, I'd suggest: `git log master@{u}..` which compares the upstream of master to your checked out branch.

If you check only new commits, you (a) don't need the exceptions file and (b) can change gitlogcheck rules more easily in the future.

mrigger:
Yes, I agree that `git log master@{u}..` is better than the approach that I initially implemented. I am pushing a new version.

mx.sulong/mx_sulong.py
```diff
@@ -32,6 +34,11 @@
 
 _dragonEggPath = _toolDir + 'tools/dragonegg/dragonegg-3.2.src/dragonegg.so'
 
+gitLogFile = _mx + 'gitlogfile.cfg'
```
woess: This variable is now unused, isn't it?

mrigger: Yes, thanks! I removed it.

PR #174, opened 2016-04-20
Do not execute destructors for multiple program executions

mrigger:
The destructors should only be executed once after the last program run. If they get executed before the start of a new program run they deallocate global variables that are not allocated in the static constructors, but in the parser.


rschatz:
Are you sure you're allowed to do this?
I would expect that you either have to execute both the destructors *and* the constructors, or execute *none* of them. Just re-executing the constructors without executing the destructors in between seems wrong.

mrigger:
I had it like this before, and since multiple execution runs are not tested as part of the Travis gate, I only figured out now that executing both does not work. I think we have to discriminate two types of destructors to make this clean:
1. destructors functions (`__attribute__((destructor))`) of the program that need to be called before the constructors
2. other destructors that free global variables that are allocated in the parser

Since destructor functions (1.) are not implemented yet, this approach works for the current implementation. Execution also behaves as expected when I locally execute my benchmark suite.

rschatz:
Ok. We just have to make sure we don't forget about it ;)

In general, we should probably think about the "proper" initialization and destruction sequence that needs to be done when executing a program multiple times, so the program will not see dirty state from the previous execution.

PR #177, opened 2016-04-20
C format checks

mrigger:
This change introduces a `clang-format` command to validate the format of the `include/truffle.h` file. It would make sense to also apply `clang-format` to our test cases in the future.


rschatz:
Nice ;)

Did you thest that it will actually fail in the gate if there is a formatting error?

mrigger:
I tested locally that it fails for formatting errors.

PR #180, opened 2016-04-21
Improve the performance of function descriptor lookup from the function address

mrigger:
Improves the performance of the function descriptor lookup by not looking it up in an array but constructing a new one.


chrisseaton:
Can you update this?

mrigger:
Yes, I still have to fix an issue to pass the gate.

PR #181, opened 2016-04-21
Remove c substitutions

mrigger:
This change removes the substitutions for trigonometric C functions, since calls to the native standard library implementations are faster (a factor of two on my machine).


chrisseaton:
> the Graal intrinsifications for some trigonometric functions in {@link AMD64MathSubstitutions} are still twice as slow as their C counterparts

Why on earth is this? I thought they were just direct FPU instructions.

mrigger:
I found this explanation on [Stackoverflow](http://stackoverflow.com/questions/2284860/how-does-c-compute-sin-and-other-math-functions):  

> It turns out that the most optimized math libraries on x86 use SSE software implementations for sin and cos that are faster than the hardware instructions on the FPU. Simpler, more naive libraries tend to use the fsin and fcos instructions

rschatz:
We should really have a better intrinsification for this in Graal. Calling FPU instructions is really slow, especially since there is no direct move instruction from XMM registers to FPU registers, so the argument and result have to move through the stack...

Until we have better intrinsics in GraalVM, I guess just calling the C standard library is an acceptable second-best solution. Depending on how complicated the intrinsic code ends up, it might even be *the* best solution.

PR #183, opened 2016-04-22
Provide a basic implementation for implicit foreign array access

grimmerm:
This commit provides a basic implementation to directly access foreign arrays using C syntax.
E.g.
``` C
int *foreign = (int*) truffle_import("arr");
int value = foreign[2];
```


projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
```diff
@@ -1164,7 +1164,8 @@ private LLVMExpressionNode getConstantElementPtr(LLVMExpressionNode baseAddress,
         }
         if (currentOffset != 0) {
             LLVMExpressionNode oneValueNode = factoryFacade.createLiteral(1, LLVMBaseType.I32);
```
mrigger: Can we instead of introducing a new field in the `NodeFactoryFacade` just adapt the value of the `oneValueNode` to be `currentOffset/stride`?

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
```diff
@@ -869,21 +869,21 @@ private LLVMExpressionNode getElementPtr(LLVMExpressionNode baseAddress, Type ba
...
                         case I64:
-                            constantNode = factoryFacade.createLiteral(1L, LLVMBaseType.I64);
+                            constantNode = factoryFacade.createLiteral(constantIndex.longValue(), LLVMBaseType.I64);
                             break;
                         default:
                             throw new AssertionError();
                     }
-                    currentAddress = factoryFacade.createGetElementPtr(getLLVMType(type), currentAddress, constantNode, indexedTypeLength);
+                    currentAddress = factoryFacade.createGetElementPtr(getLLVMType(type), currentAddress, constantNode, stride);
                 }
```
mrigger: The change in `LLVMVisitor` ignores that the GEP instruction can also index structs since it only takes in account the stride computed by the first element and the value of the GEP index. `LLVMTypeHelper.goIntoTypeGetLengthByte` computed the offset into a struct or array. Maybe you can introduce a new method in LLVMTypeHelper that returns a tuple (stride, index) that returns the the stride and index for arrays, but uses `LLVMTypeHelper.goIntoTypeGetLengthByte` and constant 1 otherwise?

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
```diff
@@ -875,21 +875,28 @@ private LLVMExpressionNode getElementPtr(LLVMExpressionNode baseAddress, Type ba
                 LLVMExpressionNode valueRef = visitValueRef(currentRef, type);
                 currentAddress = factoryFacade.createGetElementPtr(getLLVMType(type), currentAddress, valueRef, indexedTypeLength);
             } else {
-                int indexedTypeLength = LLVMTypeHelper.goIntoTypeGetLengthByte(currentType, constantIndex);
+                LLVMBaseType llvmType = LLVMTypeHelper.getLLVMType(currentType);
+                int stride;
+                if (llvmType == LLVMBaseType.ARRAY || llvmType == LLVMBaseType.ADDRESS) {
+                    stride = LLVMTypeHelper.getByteSize(currentType.getContainedType(0));
```
mrigger: I think that the following works:
stride = LLVMTypeHelper.goIntoTypeGetLengthByte(currentType, 1);

PR #184, opened 2016-04-22
Branch probability injection in switch nodes

mrigger:
This change implements nodes that inject branch probabilities for switch instructions, similarly to branch and select instruction nodes.

PR #186, opened 2016-04-24
Annotation fixes in switch node

mrigger:
This change addresses comments in #184 and fixes some wrong and misleading annotations.

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/control/LLVMSwitchNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/control/LLVMSwitchNode.java
@@ -102,7 +102,6 @@ public abstract class LLVMSwitchNode extends LLVMTerminatorNode {
             return DEFAULT_LABEL_INDEX;
         }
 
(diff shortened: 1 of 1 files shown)


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/control/LLVMSwitchNode.java
```diff
@@ -130,6 +129,7 @@ public LLVMI8ProfilingSwitchNode(LLVMI8Node cond, LLVMI8Node[] cases, int[] succ
         }
 
         @Override
+        @ExplodeLoop
         boolean profile(int i, boolean value) {
             return profiles[i].profile(value);
         }
```
chrisseaton: This use of explode loop still doesn't make sense. There's no loop here! Maybe you have a misunderstanding about what the annotation does? For example it's not transient to called methods as well, if that's what you think.

mrigger: Yes, I agree that it does not make sense. I am not sure how I came up with using the annotation here.

PR #198, opened 2016-04-26
Fix inconsistent suite formatting

chrisseaton:


mrigger:
Are there no automatic checks for that?

chrisseaton:
Apparently not.

mrigger:
It seems we just check other aspects such as trailing whitespaces.

PR #199, opened 2016-04-26
Add bitcode support to Sulong

crbb:
A new project has been added to the Sulong suite to contain a bitcode decoder library to complement the IR Editor based parser. The project relies on an external jar file for the decoding and the com.oracle.truffle.llvm.frontend project implements the visitor interface used in the model generated by the library. The library supports LLVM 3.2 used by Sulong but also LLVM 3.8 which is the current release. This allows current LLVM projects to be interfaced with Sulong. It also allows current LLVM language frontends and backend targets. There is also the potential to support other LLVM versions (e.g. Google's interpretation) and metalanguages that use metadata on IR instructions.


mrigger:
Thanks for the great contribution! I'll try to finish reviewing the change as soon as possible.

projects/com.oracle.truffle.llvm.frontend/com.oracle.truffle.llvm.frontend
```diff
@@ -0,0 +1,114 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mrigger: Did you accidentally include this file?

projects/com.oracle.truffle.llvm.frontend/src/com/oracle/truffle/llvm/frontend/bc/LLVMBitcodeHelper.java
```diff
@@ -0,0 +1,973 @@
...
+                return LLVMStoreNodeFactory.LLVMI16ArrayLiteralNodeGen.create(values.toArray(new LLVMI16Node[nrElements]), baseTypeSize, target);
+            case I32:
+                return LLVMStoreNodeFactory.LLVMI32ArrayLiteralNodeGen.create(values.toArray(new LLVMI32Node[nrElements]), baseTypeSize, target);
+            case I64:
+                return LLVMStoreNodeFactory.LLVMI64ArrayLiteralNodeGen.create(values.toArray(new LLVMI64Node[nrElements]), baseTypeSize, target);
+            case FLOAT:
+                return LLVMStoreNodeFactory.LLVMFloatArrayLiteralNodeGen.create(values.toArray(new LLVMFloatNode[nrElements]), baseTypeSize, target);
+            case DOUBLE:
+                return LLVMStoreNodeFactory.LLVMDoubleArrayLiteralNodeGen.create(values.toArray(new LLVMDoubleNode[nrElements]), baseTypeSize, target);
+//            case X86_FP80:
```
mrigger: That is not supported yet?

projects/com.oracle.truffle.llvm.frontend/com.oracle.truffle.llvm.frontend
crbb: Yes, I did. I think it is the file automatically generated by the mx script. I am not sure if .gitignore needs to be updated to prevent its inclusion.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
```diff
@@ -66,11 +66,9 @@ public static LLVMContext create(NodeFactoryFacade facade, LLVMOptimizationConfi
```
mrigger: I think that this commit is obsolete with #182.

projects/com.oracle.truffle.llvm.frontend/src/com/oracle/truffle/llvm/frontend/bc/LLVMBitcodeHelper.java
crbb: I have only looked at the su-tests-sulong so far so for integration. The half, x86_fp80, fp128 and ppc_fp128 data types are supported in the decoder but as I have had no need to integrate it or test them, the fp80 remains commented out.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
crbb: You are correct.

mrigger:
Great code quality!

I would suggest that we break the PR into smaller parts, that we can immediately merge, and others that we can still discuss further. This would speed up reviewing and would prevent that the development on the main branch diverges too much from this PR. 

I would suggest the following PRs that we could directly merge:

- the abstract parser interface (Add interface to abstract parser result, Add bitcode support to Sulong)
- the commit "Fix inconsistency of GGC version used in test tools"
- remove IREditor dependencies not related to the vector implementations (Add factory method to remove dependency on IREditor and runtime, Add factory method to create branch without runtime dependency, Add factory method to root factory to remove runtime dependency)
- the vector implementations (Add nodes to support vector operations used by clang, Fix protected method names for checkstyle, Add factory method to create vector comparisons, Add factory method to create vector casts, Remove dependency on IREditor from vector literals, Add factory methods for load/store without IREditor dependency, Add factory method to create vector select nodes, Add factory method to insert element without runtime dependency)

Would that be okay for you?

projects/com.oracle.truffle.llvm.frontend/src/com/oracle/truffle/llvm/frontend/bc/LLVMBitcodeHelper.java
mrigger: I would prefer to have no code that is commented out in the Sulong code base. I agree with [Clean Code p. 68](vitoex.googlecode.com/svn/trunk/Read/Clean%20Code.pdf) that it's usually not a good idea to keep it.

mrigger:
We integrated this change with #210.

PR #200, opened 2016-04-26
Fix gcc dependency in sulong tests and add assembler

crbb:
Sulong specifies gcc-4.6 (with fortran) as a dependency. During su-tests-sulong the gcc alias is used resulting in failed tests if a matching library set is not installed. Change it to gcc-4.6 to match dependency. Also this commit takes the opportunity to create a Java interface to the LLVM assembler (llvm-as) in the tools project.


projects/com.oracle.truffle.llvm.tools/src/com/oracle/truffle/llvm/tools/LLVM_AS.java
```diff
@@ -0,0 +1,46 @@
...
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.tools;
+
+import java.io.File;
+
+import com.oracle.truffle.llvm.tools.util.ProcessUtil;
+
+public final class LLVM_AS {
```
mrigger: Could you rename the class to `LLVMAs` or `LLVMAssembler`? Although there are some classes in the Sulong base with an underscore (the vararg intrinsics such as `LLVMX86_64BitVAStart`) I think that class names without underscores would fit better to the rest of the code base. If you do not agree because `llvm-as` is a special case, I can also merge the change like this.

PR #201, opened 2016-04-26
Add nodes to support vector cast, compare and select

crbb:


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/op/compare/LLVMI32VectorCompareNode.java
```diff
@@ -0,0 +1,201 @@
...
+import com.oracle.truffle.llvm.types.vector.LLVMI1Vector;
+import com.oracle.truffle.llvm.types.vector.LLVMI32Vector;
+
+@NodeChildren({
+    @NodeChild(value = "addressNode", type = LLVMAddressNode.class),
+    @NodeChild(value = "leftNode", type = LLVMI32VectorNode.class),
+    @NodeChild(value = "rightNode", type = LLVMI32VectorNode.class)})
+public abstract class LLVMI32VectorCompareNode extends LLVMI1VectorNode {
+
+    protected LLVMI1Vector executeI1VectorBody(LLVMAddress target, LLVMI32Vector left, LLVMI32Vector right) {
```
mrigger: I just noticed that you need a `@ExplodeLoop` here and at the other execute<type>VectorBody methods.

mrigger: @chrisseaton Is that right?

PR #210, opened 2016-04-28
Add the new project for bitcode decoding

crbb:
A projects has been created that contains the initial implementation of the interface between LLVM bitcode and Sulong. The project is not used but is in place for refactoring, unification with the node factory facades and integration into the LLVM class.


chrisseaton:
@crbb can you update the branch, and then I will merge.

PR #215, opened 2016-04-29
Revert branch probs remove

mrigger:
I previously assumed that branch probability profiling for successors in #211 made the profiling in switches, selects, and indirect branches redundant. However, removing them made most benchmarks slower. For example, removing the indirect branch profiling for indirect branches alone makes the whetstone benchmark 10% slower. Hence, this change reverts the removal of the branch profiling in the switch, br, and select nodes.


lukasstadler:
hm, the problem might be that public double getBranchProbability(int successorIndex) only does transferToInterpreter if the total execution count is zero, but not if the execution count of the specific successor is zero. Since normal counting profiles to this, I'd suggest to try this before reverting everything.

mrigger:
Yes, thanks! Unfortunately the comment in the code only pretended to deopt if the count of a successor is zero, and did not implement it. ;)

Fixed in #217

mrigger:
It seems that this is still a problem for switch, since the comparisons have all the same (wrong) probability.

Issue #196, opened 2016-04-25
Interop can't handle TruffleObject as LLVMAddress


chrisseaton:
(#195 needed to see this problem)

My Ruby C extension API implementation has this code:

```c
static void *ruby_cext;

__attribute__((constructor))
void truffle_ruby_load() {
  ruby_cext = truffle_import("ruby_cext");
}
```

The import gives a `TruffleObject` and the assign to `ruby_cext` tries to execute `LLVMAddressStoreNode`, which executes `LVMAddressReadNode`, which tries to cast the `TruffleObject` to `LLVMAddress`.

```
java.lang.ClassCastException: com.oracle.graal.truffleom.a.c cannot be cast to com.oracle.truffle.llvm.types.LLVMAddress
	at com.oracle.truffle.llvm.nodes.impl.vars.LLVMReadNodeFactory$LLVMAddressReadNodeGen.executePointee(LLVMReadNodeFactory.java:442)
	at com.oracle.truffle.llvm.nodes.impl.memory.LLVMStoreNodeFactory$LLVMAddressStoreNodeGen.executeVoid(LLVMStoreNodeFactory.java:358)
	at com.oracle.truffle.llvm.nodes.impl.base.LLVMBasicBlockNode.executeGetSuccessorIndex(LLVMBasicBlockNode.java:63)
	at com.oracle.truffle.llvm.nodes.impl.others.LLVMBlockNode$LLVMBlockControlFlowNode.executeGeneric(LLVMBlockNode.java:71)
	at com.oracle.truffle.llvm.nodes.impl.func.LLVMFunctionStartNode.execute(LLVMFunctionStartNode.java:63)
	at com.oracle.graal.truffle.OptimizedCallTarget.callProxy(OptimizedCallTarget.java:503)
	at com.oracle.graal.truffle.OptimizedCallTarget.callRoot(OptimizedCallTarget.java:365)
	at com.oracle.graal.truffle.OptimizedCallTarget.callBoundary(OptimizedCallTarget.java:351)
	at com.oracle.graal.truffle.OptimizedCallTarget.doInvoke(OptimizedCallTarget.java:339)
	at com.oracle.graal.truffle.OptimizedCallTarget.callDirect(OptimizedCallTarget.java:254)
	at com.oracle.graal.truffle.OptimizedDirectCallNode.callProxy(OptimizedDirectCallNode.java:70)
[...]

PR #219, opened 2016-04-29
Add truffle_import_cached in truffle.h

grimmerm:
This function allows us to import a foreign value (see also truffle_import function).
In contrast to truffle_import,  truffle_import_cached caches the imported value internally.
This caching guarantees good performance because we save the foreign value look-up.

Truffle_import_cached allows you to import a foreign value and keep is as a function local (rather than a global variable) without a loss in performance (which would be introduced by importing a foreign value every time truffle_import is called). 

For example:

Rather than writing 
```C
static void *ruby_cext;

__attribute__((constructor))
void truffle_ruby_load() {
  ruby_cext = truffle_import("ruby_cext");
}
```

one can now do

```C
__attribute__((constructor))
void truffle_ruby_load() {
  void *ruby_cext = truffle_import_cached("ruby_cext");
}
```
.

This should solve #196.

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleImportCached.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleImportCached.java
@@ -32,6 +32,7 @@ package com.oracle.truffle.llvm.nodes.impl.intrinsics.interop;
 import com.oracle.truffle.api.dsl.Cached;
 import com.oracle.truffle.api.dsl.NodeChild;
 import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.api.dsl.internal.SuppressFBWarnings;
 import com.oracle.truffle.api.interop.TruffleObject;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.llvm.nodes.impl.base.LLVMAddressNode;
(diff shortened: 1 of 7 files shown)


chrisseaton:
I don't see the need for a separate name for the cached variant. Why not make all imports cached?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleImportCached.java
```diff
@@ -0,0 +1,103 @@
...
+            if (this.name.equals(n)) {
+                return value;
+            } else {
+                return next.execute(n);
+            }
+        }
+
+    }
+
+}
```
chrisseaton: Could this not be implemented using the DSL's `@Cached`?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleImportCached.java
```diff
@@ -0,0 +1,103 @@
...
+    private static final class ResolvedCacheEntry extends CacheEntry {
+
+        private final String name;
+        private final LLVMTruffleObject value;
+        @Child private CacheEntry next;
+
+        public ResolvedCacheEntry(String name, LLVMTruffleObject value, CacheEntry next) {
+            this.name = name;
+            this.value = value;
+            this.next = insert(next);
```
woess: you shouldn't have to insert in the constructor

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleImportCached.java
```diff
@@ -0,0 +1,103 @@
...
+            TruffleObject foreignObject = (TruffleObject) LLVMLanguage.INSTANCE.getEnvironment().importSymbol(name);
+            LLVMTruffleObject value = new LLVMTruffleObject(foreignObject);
+
+            Node n = this;
+            while (!(n instanceof LLVMTruffleImportCached)) {
+                n = n.getParent();
+            }
+            LLVMTruffleImportCached head = (LLVMTruffleImportCached) n;
+            head.cache = insert(new ResolvedCacheEntry(name, value, head.cache));
+            head.adoptChildren();
```
woess: I believe this should be `head.insert(...)` - probably the reason why you added `adoptChildren`?

grimmerm:
@chrisseaton, the thing with always caching is that we would miss an updated exported symbol.

e.g.
JS: export("foo", a)
C: import("foo") // imports a
JS: export("foo", b)
C: import("foo") // would return the cached a object

do we really want that?

chrisseaton:
We should probably implement invalidation within the Truffle API. When you import an object, you could get the object but also an `Assumption` which is invalidated when that name is replaced with a new object. The Truffle API would manage that invalidation.

PR #220, opened 2016-04-29
Remove randomly failing test case

mrigger:
This test case has recently been failing nondeterministically. To not block other PRs, this change removes the test case.


chrisseaton:
Can you exclude rather than remove it?

mrigger:
Okay, I excluded the file by renaming it.

I usually try to remove test cases that are not executed due to the same reason as I remove commented-out code. If nobody fixes or investigates the test case, then it will soon be forgotten and after a month or so nobody will ever dare to touch it. If we remove it, we can still just get the test case from the Git history.

chrisseaton:
For example in Ruby we see the number of excluded test cases, so it's something you get reminded about every time you run tests and we can track, but maybe that's not the case here with your test setup.

PR #226, opened 2016-05-02
Add test case for ordered comparisons with NaN

mrigger:


mrigger:
I added this change to #229.

PR #227, opened 2016-05-02
Add @ValueType annotation to the Sulong runtimes types

mrigger:


chrisseaton:
What does this do?

lukasstadler:
it makes types behave like boxing types you created via valueOf, i.e., they don't guarantee consistency in terms of object identity.
("Integer.valueOf(x) == Integer.valueOf(x)" may be true or false - "new Integer(x) == new Integer(x)" is guaranteed to be false)

lukasstadler:
a compiler can use this to optimize more aggressively.

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMFunctionDescriptor.java
```diff
@@ -45,6 +46,7 @@
 import com.oracle.truffle.api.nodes.RootNode;
 import com.oracle.truffle.llvm.types.memory.LLVMStack;
 
+@ValueType
 public final class LLVMFunctionDescriptor implements TruffleObject, Comparable<LLVMFunctionDescriptor> {
```
lukasstadler: I doubt that this is correct here, because this type is potentially stored in Maps and passed around to unsuspecting foreign languages. This can be a problem since ValueType breaks some of the most basic contracts about the behavior of Java objects.

mrigger: How is this case different from `LLVMAddress`? Both are mainly runtime data structures. Also, `LLVMFunctionDescriptor` implements both `hashCode` and `equals` based on `functionId` field.

mrigger:
I modified the commit message to explain the motivation for this change, based on Lukas' answer.

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMFunctionDescriptor.java
lukasstadler: since it's a TruffleObject, I assumed that it gets handed out to other languages.

PR #228, opened 2016-05-02
Add first part of inline assembler support

PrinzKatharina:


projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/inlineassembly/LLVMInlineAssemblyTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Test
+    public void test002() {
+        String file = "inlineassembly002";
+        Assert.assertEquals(42, run(file, null, null));
+    }
+
+    private static int run(String fileName, TruffleObject foreignObject, String foreignObjectName) {
+        Builder builder = PolyglotEngine.newBuilder();
+        builder.globalSymbol(foreignObjectName, foreignObject);
```
mrigger: You do not need the `foreignObject` and `foreignObjectName` arguments here. I think it would be best if you used the [TestGCCSuite](https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/TestGCCSuite.java) format. For now it is sufficient to remove the parameters.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/inlineassembly/LLVMInlineAssemblyTest.java
```diff
@@ -0,0 +1,83 @@
...
+        String file = "inlineassembly001";
+        Assert.assertEquals(42, run(file, null, null));
+    }
+
+    @Test
+    public void test002() {
+        String file = "inlineassembly002";
+        Assert.assertEquals(42, run(file, null, null));
+    }
+
```
mrigger: You checked in some files `inlineassembly003.c` and `inlineassembly004.c` that are not executed. Can you remove the files or execute them here?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMInlineAssemblyRootNode.java
```diff
@@ -0,0 +1,53 @@
...
+import com.oracle.truffle.api.nodes.RootNode;
+import com.oracle.truffle.api.source.SourceSection;
+import com.oracle.truffle.llvm.nodes.base.LLVMExpressionNode;
+
+public class LLVMInlineAssemblyRootNode extends RootNode {
+
+    @Child private LLVMExpressionNode node;
+
+    public LLVMInlineAssemblyRootNode(@SuppressWarnings("rawtypes") Class<? extends TruffleLanguage> language, SourceSection sourceSection, FrameDescriptor frameDescriptor, LLVMExpressionNode node) {
+        super(language, sourceSection, frameDescriptor);
```
mrigger: You do not need to pass the `language` to the constructor. You can just use `super(LLVMLanguage.class, sourceSection, frameDescriptor)` such as in [LLVMGlobalRootNode](https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMGlobalRootNode.java)

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/asm/LLVMAsmSublNode.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package com.oracle.truffle.llvm.nodes.impl.asm;
+
+import com.oracle.truffle.api.dsl.NodeChild;
+import com.oracle.truffle.api.dsl.NodeChildren;
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.llvm.nodes.impl.base.integers.LLVMI32Node;
+
+@NodeChildren({@NodeChild("leftNode"), @NodeChild("rightNode")})
+public abstract class LLVMAsmSublNode extends LLVMI32Node {
```
mrigger: I think it would be clearer if we renamed the classes in this package `LLVMx86<...>Node` since the nodes are specific to x86. However, you can do that in a separate PR.

mrigger:
Great, thanks for the PR! I'm already looking forward to execute inline assembler on Sulong. I found some minor things that you can improve which I noted as inline comments. You accidentally committed a file `projects/com.oracle.truffle.llvm.test/inlineassemblytests/.DS_Store` that you should remove. I will update the `.gitignore` file to prevent this in future.

mx.sulong/suite.py
```diff
@@ -357,11 +357,25 @@
       "license" : "BSD-new",
     },
 
+    "com.oracle.truffle.llvm.assemblerX86" : {
```
chrisseaton: If this is 64 bit x86, then I think Oracle usually uses the name 'AMD64'. Might be good to use that for consistency?

mrigger: Maybe we can rename it to `com.oracle.truffle.llvm.asm.amd64`? Then it would be consistent with `com.oracle.graal.asm.amd64`.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/asm/LLVMAsmSublNode.java
mrigger: or better LLVMAMD64<...>Node?

PR #229, opened 2016-05-02
Simplification of ordered double comparisons

mrigger:
This change removes isOrdered checks in ordered double comparisons based on the fact that NaN comparisons always result in false. It also adds some test cases for ordered double comparisons.


gilles-duboscq:
Did you check if you can use the same things of the unordered checks?
For example transforming
```java
!areOrdered(val1, val2) || val1 > val2
```
into
```java
!(val1 <= val2)
```
I think it should work.

mrigger:
Yes, but I want to address this in an additional PR.

PR #231, opened 2016-05-02
Add truffle_read_string intrinsic to turn a char* into a String

chrisseaton:


chrisseaton:
Used for example like this https://github.com/jruby/jruby/compare/truffle-c-exts?expand=1#diff-c14b15a0d092eb68d1ea3908676615e5R188. Otherwise nobody knows to turn the address into something that Ruby can see is a string.

mrigger:
Can we have a test case for this, so we do not accidentally break the new function?

chrisseaton:
Added a test.

grimmerm:
I approve the change.

Issue #224, opened 2016-05-01
Lower the Travis gate execution time


mrigger:
Currently, a Travis gate run takes around 30 to 40 minutes (1:20 to 1:40 total time). I think that it should be shorter to not become a development bottleneck on days on which several people actively develop and submit PRs.

The most time is spent in the `mx su-travis1` and `mx su-travis2` steps. Both steps are supposed to mainly execute test cases. However, a major portion is actually spent in compiling the native (modified HotSpot) project. We have to build it since we have to execute the test cases with Graal. We directly rely on the Graal Native Function Interface for executing native calls and thus cannot use a standard JVM. Breaking the test cases into several independent commands does not scale, since we have to compile the native project for each of this steps. 

One idea to solve this is to use a Travis cache (or another cache) for the native project. Instead of building the project, we could directly download the built native project. This would reduce the built time and would also allow us to break the test cases into smaller parts. One drawback of this solution is that we have to upload a new version whenever the native project is updated. If nobody has a better idea, I will still try to follow this approach.

mrigger:
Using binary JVMCI dependencies (see #234) and Travis caches (see #237) reduced the gate execution time to around 25 minutes (around 1:00 total time).

PR #234, opened 2016-05-03
Binary jvmci dependency

mrigger:
This change addresses #224 by not building the jvmci native from source but downloading the binary from https://curio.ssw.jku.at/nexus/content/repositories/snapshots to reduce the Travis gate time.


sanzinger:
LGTM.

PR #235, opened 2016-05-03
Simplify ordered float comparisons and add test cases

mrigger:


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/op/compare/LLVMFloatCompareNode.java
```diff
@@ -54,42 +54,71 @@ private static boolean areOrdered(float... vals) {
     public abstract static class LLVMFloatOltNode extends LLVMFloatCompareNode {
         @Specialization
         public boolean executeI1(float val1, float val2) {
-            return areOrdered(val1, val2) && val1 < val2;
+            if (val1 < val2) {
+                assert areOrdered(val1, val2);
+                return true;
+            } else {
+                return false;
+            }
```
chrisseaton: I don't know if it's worth profiling at this point? Or is this compare node only ever used in cases which will profile anyway, like conditional jumps?

mrigger: If the comparison is used in an branch, then the branch probability profiling that is already injected in the successor selection should suffice. However, the value could also be stored into the frame or somewhere else, and then used at a different location. Do you think that branch probability injection could improve performance in this case?

/cc @lukasstadler @gilles-duboscq

chrisseaton: In Ruby in general if we have an if statement that we don't expect to always be a constant in code that is being PEd, we will use a profile.

gilles-duboscq: Isn't this somehow a larger design question?
Do you profile your conditions/logic nodes or do you profile the things that use them (such as branching control-flow nodes)

woess: IMHO, the latter (branching control flow). Profiling the input condition of an `if` is a convenient way to do that but the profile still belongs to the `if` (and its node), not the condition.

PR #236, opened 2016-05-03
Disable sudo in the Travis configuration to switch to the container-based infrastructure

mrigger:


mrigger:
This change is obsolete since we can now use Travis caches without using the container-based infrastructure (see #237).

PR #240, opened 2016-05-04
Work to support Ruby cexts

chrisseaton:


projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMFunctionDescriptor.java
```diff
@@ -228,10 +229,13 @@ protected ForeignCallNode() {
...
         public Object execute(VirtualFrame frame) {
-            assert ForeignAccess.getArguments(frame).isEmpty();
             final LLVMFunctionDescriptor function = (LLVMFunctionDescriptor) ForeignAccess.getReceiver(frame);
             final CallTarget callTarget = getCallTarget(function);
-            return callNode.call(frame, callTarget, new Object[]{stack.getUpperBounds()});
+            final List<Object> arguments = ForeignAccess.getArguments(frame);
+            final Object[] packedArguments = new Object[1 + arguments.size()];
+            packedArguments[0] = stack.getUpperBounds();
+            System.arraycopy(arguments.toArray(), 0, packedArguments, 1, arguments.size());
+            return callNode.call(frame, callTarget, packedArguments);
```
mrigger: It's probably not relevant for now but if the `function` is a native function or has a struct return type (see [LLVMVisitor](https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java#L661)) then the calling convention is different.

chrisseaton: Added

PR #246, opened 2016-05-11
Test against JRuby

chrisseaton:
I'm not sure if it's your job to test against JRuby or JRuby's job to test against Sulong, but building Sulong is more complex, so I'll reuse that by doing it in this repository.

This checks that JRuby can use this version of Sulong to build and run C extensions.


chrisseaton:
@mrigger I need to build `sulong.jar` for this test to work, but can't figure out how to do it. Do you have any idea?

mrigger:
In the [travis2](https://github.com/graalvm/sulong/blob/9553fa123b48d9efb711e6b260ecfa1fcb979883/mx.sulong/mx_sulong.py#L132) function we build all projects including the distributions with the following lines:
```
    tasks = []
    with Task('BuildHotSpotGraalServer: product', tasks) as t:
        if t: buildvms(['-c', '--vms', 'server', '--builds', 'product'])
```

You should be able to find the distribution in the `build` directory. To include the distribution in the classpath, you can use [`getClassPathOptions`](https://github.com/graalvm/sulong/blob/9553fa123b48d9efb711e6b260ecfa1fcb979883/mx.sulong/mx_sulong.py#L630)

mx.sulong/mx_sulong.py
```diff
@@ -512,6 +520,22 @@ def runCompileTestCases(args=None):
...
+    """tests that JRuby can use this version of Sulong to compile and run C extensions"""
+    rubyUrl = 'http://github.com/jruby/jruby.git'
+    rubyBranch = 'truffle-head'
+    suitesDir = os.path.abspath(join(_suite.dir, '..'))
+    jrubyDir = join(suitesDir, 'jruby')
+    if os.path.isdir(jrubyDir):
+        mx.run(['git', 'checkout', rubyBranch], cwd=jrubyDir)
+        mx.run(['git', 'pull'], cwd=jrubyDir)
+    else:
+        mx.run(['git', 'clone', rubyUrl], cwd=suitesDir)
```
mrigger: Can we use `--depth=1` to speed up the tests?

chrisseaton: It looks like a shallow clone of a specific branch is tricky to do http://stackoverflow.com/questions/4929688/how-to-shallow-clone-a-single-branch-in-git

mrigger: Okay, then let's stay with the full clone.

mrigger:
Do you think that bugs in JRuby could block Sulong development? If yes we could use Travis' [allow failures feature](https://docs.travis-ci.com/user/customizing-the-build/) for the integration testing.

chrisseaton:
I'd like to give it a go as a required pass to start with. If we find it does break unrelated to changes in Sulong frequently then we can reconsider, but if it hardly ever breaks then why not make it required?

mrigger:
Okay, I'm fine with that.

chrisseaton:
This is failing due to this weird error:

```
Caused by: java.lang.ClassCastException: com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionInterfaceAccess cannot be cast to com.oracle.nfi.api.NativeFunctionInterfaceAccess
```

https://travis-ci.org/graalvm/sulong/builds/129862619#L3620

Any ideas?

mx.sulong/mx_sulong.py
chrisseaton: I think Travis has some kind of caching or direct link to GitHub, because this clone only takes a second or two anyway.

mrigger:
From the error message I would say that `graal-truffle-hotspot.jar` is not on the class path. However, I see it twice in the `JAVACMD`. Could that be the problem?

woess:
Probably `com.oracle.nfi.*` should be on the bootclasspath. You can also try with `-XX:-UseJVMCIClassLoader`.

woess:
BTW, It's not very obvious that travis3 is testing against ruby. Should we give it a more descriptive name, like travis-ruby? Or is the order important?

chrisseaton:
Done.

PR #250, opened 2016-05-13
Qsort

mrigger:
This change adds a naive qsort implementation that is loaded for each Sulong execution.


chrisseaton:
What's it for?

mrigger:
For supporting qsort with a Truffle function comparator. If we do not have the qsort implementation as a Truffle AST then we could only supply native functions as a comparator callback.

PR #256, opened 2016-05-16
Add basic debugging support to Sulong

flortsch:
This PR adds basic debugging support to Sulong.

It creates a new instrumentable LLVMStatementNode class with LLVMTerminatorNode and LLVMWriteNode as subclasses. Furthermore it adopts constructor changes in the corresponding node factories. This PR adds the ProvidedTags annotation to the LLVMLanguage class and adds dummy source sections to statement nodes in LLVMVisitor. Finally the PolyglotEngine in the LLVM class is extended with some very simple event handlers.


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMTerminatorNode.java
```diff
@@ -29,25 +29,46 @@
...
+ * generated from super class {@link LLVMStatementNode}. Therefore this class is also {@link Instrumentable}.
+ *
  * @see <a href="http://llvm.org/docs/LangRef.html#terminator-instructions">terminator
  *      instructions</a>
  */
-public abstract class LLVMTerminatorNode extends LLVMNode {
+@Instrumentable(factory = LLVMTerminatorNodeWrapper.class)
+public abstract class LLVMTerminatorNode extends LLVMStatementNode {
 
-    @CompilationFinal private final int[] successors;
```
mrigger: Is it an accident that this field is no longer compilation final?

projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
```diff
@@ -241,10 +244,27 @@ public static int executeMain(String codeString, Object... args) {
...
-        engineBuilder.config(LLVMLanguage.LLVM_IR_MIME_TYPE, LLVMLanguage.LLVM_SOURCE_FILE_KEY, fileSource);
-        PolyglotEngine vm = engineBuilder.build();
+        Builder builder = PolyglotEngine.newBuilder();
+        builder.config(LLVMLanguage.LLVM_IR_MIME_TYPE, LLVMLanguage.MAIN_ARGS_KEY, args);
+        builder.config(LLVMLanguage.LLVM_IR_MIME_TYPE, LLVMLanguage.LLVM_SOURCE_FILE_KEY, fileSource);
+        builder.onEvent(new EventConsumer<ExecutionEvent>(ExecutionEvent.class) {
+
+            @Override
+            protected void on(ExecutionEvent executionEvent) {
+                System.out.println(executionEvent);
```
mrigger: Can we remove the two `System.out.println` for integration?

mrigger:
The changes look good! We can integrate the change after you addressed the two comments and when the changes pass the gate.

projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
flortsch: Sure.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMTerminatorNode.java
flortsch: I thought that the CompilationFinal annotation is used for fields which should be treated as final by Truffle during runtime while the field actually is not marked final. Since the field is marked final I thought the annotation would be obsolete, but correct me if I am wrong. I did the same thing for the source section field in LLVMStatementNode class.

mrigger: The Java `final` just specifies that the reference is final. You can still modify the array content, e.g., do something like `successors[3] = 0;`. The `@CompilationFinal` annotation tells the compiler that also the array content is final and is thus not redundant.

It should not be a problem to not have the source section in `LLVMStatementNode` `@CompilationFinal` since accesses to it are not on the fast path anyway.

flortsch: Oh okay, I wasn't aware of that the annotation is used, when content doesn't change. I will fix that.

flortsch:
I saw that my commit messages don't start with an upper case and that my branch was out of date with the base branch when I created the PR. Can this be fixed within this PR?

PR #257, opened 2016-05-17
Add a signal (dummy) implementation

mrigger:
This change adds a dummy implementation for the C standard library `signal` function. The motivation for this change is the same as for #250.


mx.sulong/libs/signal.c
```diff
@@ -0,0 +1,7 @@
+#include <stdio.h>
+
+void (*signal(int sig, void (*func)(int)))(int) {
+  printf("Sulong does not support signals, registering a signal handler over "
```
chrisseaton: `fprintf(stderr, ` perhaps?

PR #274, opened 2016-05-26
Add logical operations inline assembly

PrinzKatharina:


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/asm/LLVMAMD64InclNode.java
```diff
@@ -0,0 +1,44 @@
...
+import com.oracle.truffle.api.dsl.NodeChild;
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.llvm.nodes.impl.base.integers.LLVMI32Node;
+
+@NodeChild("valueNode")
+public abstract class LLVMAMD64InclNode extends LLVMI32Node {
+
+    @Specialization
+    protected int executeI32(int value) {
+        return (value + 1);
```
mrigger: Should `incl` and `decl` not also store the result in the register? Did you try to use `incl` or `decl` twice in an inline assembler expression?

Issue #270, opened 2016-05-23
Access to non-native variable arguments


eregon:
@mrigger @grimmerm @chrisseaton 
For Ruby C-extensions, we would like a way to access variadic function arguments without moving them to native. For instance, there is
```c
VALUE rb_ary_new_from_args(long n, ...);
```
which creates a Ruby array from the arguments.

Using `stdarg.h` and `va_list`, `va_start`, etc would move everything to native, and it seems difficult to do differently.
```c
VALUE rb_ary_new_from_args(long n, ...) {
  va_list args;
  VALUE array = rb_ary_new_capa(n);
 
  va_start(args, n);
  for (int i = 0; i < n; i++) {
    rb_ary_store(array, i, va_arg(args, VALUE)); // va_arg produces a native address here.
  }
  va_end(args);

  return array;
}
```

Should we add an intrinsic to get an argument directly from the Object[] arguments?
Maybe `truffle_get_arg(int i)`, returning an `Object`?
Or `truffle_get_args`, returning a `Object[]`?

eregon:
Fixed by #275 .

PR #275, opened 2016-05-26
Add truffle_get_arg(i) to get varargs without going to native

eregon:
See #270.

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMCallUnboxNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMCallUnboxNode.java
@@ -193,6 +193,11 @@ public abstract class LLVMCallUnboxNode {
         public int executePointee(int value) {
             return value;
         }
+
+        @Specialization
+        public long executePointee(long value) {
+            return value;
+        }
     }
 
     @NodeChild(type = LLVMExpressionNode.class, value = "functionCallNode")
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/vars/LLVMWriteNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/vars/LLVMWriteNode.java
@@ -175,6 +175,11 @@ public abstract class LLVMWriteNode extends LLVMNode {
             frame.setObject(getSlot(), value);
         }
 
+        @Specialization
+        protected void writeLong(VirtualFrame frame, long value) {
+            frame.setObject(getSlot(), value);
+        }
+
     }
 
     @NodeChild(value = "valueNode", type = LLVMFunctionNode.class)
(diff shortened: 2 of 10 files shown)


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMCallUnboxNode.java
```diff
@@ -193,6 +193,11 @@ public LLVMAddress executePointee(LLVMAddress value) {
         public int executePointee(int value) {
             return value;
         }
+
+        @Specialization
+        public long executePointee(long value) {
+            return value;
+        }
```
eregon: At this point this looks like an identity function, should we just accept `Object` ?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/vars/LLVMWriteNode.java
```diff
@@ -175,6 +175,11 @@ protected void writeString(VirtualFrame frame, int value) {
             frame.setObject(getSlot(), value);
         }
 
+        @Specialization
+        protected void writeLong(VirtualFrame frame, long value) {
+            frame.setObject(getSlot(), value);
+        }
```
eregon: (same here about identity function)

projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/LLVMRuntimeIntrinsicFactory.java
```diff
@@ -118,6 +119,7 @@
         intrinsics.put("@truffle_import", LLVMTruffleImportFactory.getInstance());
         intrinsics.put("@truffle_import_cached", LLVMTruffleImportCachedFactory.getInstance());
         intrinsics.put("@truffle_address_to_function", LLVMTruffleAddressToFunctionFactory.getInstance());
+        intrinsics.put("@truffle_get_arg", LLVMTruffleGetArgFactory.getInstance());
```
eregon: How could I make this node to be instantiated inlined, like `va_start`, so I don't need to go to the caller frame?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMCallUnboxNode.java
chrisseaton: Yes I think so.

projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/LLVMRuntimeIntrinsicFactory.java
chrisseaton: Or we could optionally pass caller frames to the callee, like we do in Ruby sometimes.

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
```diff
@@ -672,6 +672,9 @@ private LLVMNode visitFunctionCall(Callee callee, EList<Argument> args, Resolved
             String functionName = functionHeader.getName();
             if (functionName.startsWith("@llvm.")) {
                 return factoryFacade.createLLVMIntrinsic(functionName, finalArgs, containingFunctionDef);
+            } else if (functionName.equals("@truffle_get_arg")) {
+                assert finalArgs.length == 2 : finalArgs.length;
+                return factoryFacade.createTruffleGetArgIntrinsic(finalArgs[finalArgs.length - 1]);
```
eregon: @mrigger So this is one way to do it, would you like more abstractions?

mrigger: Maybe we can have a `createTruffleIntrinsic` instead of `createTruffleGetArgIntrinsic` in the facade? For now, you could just handle `@truffle_get_arg` there and return `null` in the other cases. If the method returns `null` you can just construct a normal function call node (which is resolved during runtime in the `LLVMRuntimeIntrinsicFactory`). Then, we could also migrate the other Truffle intrinsics from `LLVMRuntimeIntrinsicFactory` in a separate step.

eregon:
@mrigger Updated, how about this?
I'd rather keep it simple for now and let LLVMRuntimeIntrinsicFactory as normal calls so the new factory only need to handle inline nodes.

mrigger:
lgtm!

PR #284, opened 2016-06-04
Use the MessageResolution DSL for LLVMFunctionDescriptor

eregon:
* This allows EXECUTE messages to compile.
* Still slightly hacky in LLVMFunctionDescriptor, one possible fix would be to move LLVMFunctionDescriptor behind a facade entirely.
* Interop EXECUTE is still slow due to native to Java String and byte-by-byte comparisons (Could we detect if it's a literal by seeing it's in read-only memory or so and then just compare pointers adresses?).

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
@@ -34,7 +34,6 @@ import java.util.List;
 import java.util.Map;
 
 import com.oracle.nfi.api.NativeFunctionHandle;
-import com.oracle.truffle.api.CallTarget;
 import com.oracle.truffle.api.CompilerAsserts;
 import com.oracle.truffle.api.ExecutionContext;
 import com.oracle.truffle.api.RootCallTarget;
@@ -66,7 +65,6 @@ public class LLVMContext extends ExecutionContext {
     public LLVMContext(NodeFactoryFacade facade, LLVMOptimizationConfiguration optimizationConfig) {
         nativeLookup = new NativeLookup(facade);
         this.registry = new LLVMFunctionRegistry(optimizationConfig, facade);
-        setLastContext(this);
     }
 
     public RootCallTarget getFunction(LLVMFunctionDescriptor function) {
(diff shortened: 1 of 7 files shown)


eregon:
cc @chrisseaton

eregon:
@grimmerm Missing `@Override` annotations in generated MessageResolution causes a compilation failure: https://travis-ci.org/graalvm/sulong/jobs/135258882
Could you fix if not already done in truffle?

mrigger:
LGTM! It reduced the number of previous hacks, so the hacky LLVMFunctionDescriptor is justifiable.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
```diff
@@ -128,7 +129,7 @@ public Source getSourceFile() {
 
     // TODO No static access to this class from LLVMFunction at the moment
 
-    private static LLVMContext lastContext;
+    @CompilationFinal private static LLVMContext lastContext;
```
chrisseaton: Really this needs to be properly fixed, rather than worked around. We should get on and do that after this is merged.

chrisseaton:
Looks good. The Travis failures look easy to fix.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
eregon: Yeah this is not even used by this change anymore I think.

eregon:
The failure needs to be fixed in Truffle AFAIK, since it's on a generated file by the MessageResolution annotation processor.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
eregon: Good catch, I removed the whole field as it is now unused.

grimmerm:
@eregon Yes, this needs a fix in truffle - however, I'm surprised that we did not see these failures in truffle gate.

mrigger:
The reason that we only see these failures in the Sulong gate stems from the stricter checks. We explicitly have a `missingOverrideAnnotationForInterfaceMethodImplementation` check in [org.eclipse.jdt.core.prefs](https://github.com/graalvm/sulong/blob/master/mx.sulong/eclipse-settings/org.eclipse.jdt.core.prefs)

grimmerm:
@eregon Should be fixed in Truffle now

PR #295, opened 2016-06-08
Truncate Sulong return value in the GCC test suite to match the nativ…

mrigger:
…e process return value


projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/TestGCCSuite.java
```diff
@@ -91,4 +92,8 @@ public void test() {
         }
     }
 
+    private static int truncate(int retValue) {
+        return retValue & (UNSIGNED_BYTE_MAX_VALUE - 1);
```
woess: nitpicking: actually, the unsigned byte max value is 0xff == 255, so I'd change the constant to that.

mrigger: Thanks, I fixed it: #302.

PR #297, opened 2016-06-08
Support void main functions

mrigger:
Both Clang and GCC support void main functions, although such functions are undefined in ANSI C. This change returns 214 for main functions to imitate the behavior of Clang.

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMMainFunctionReturnValueRootNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMMainFunctionReturnValueRootNode.java
@@ -83,4 +83,20 @@ public abstract class LLVMMainFunctionReturnValueRootNode extends RootNode {
 
     }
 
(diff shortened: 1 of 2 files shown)


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMMainFunctionReturnValueRootNode.java
```diff
@@ -83,4 +83,21 @@ public Object execute(VirtualFrame frame) {
 
     }
 
+    public static class LLVMMainFunctionReturnVoidRootNode extends LLVMMainFunctionReturnValueRootNode {
+
+        // Clang uses 214 and GCC 237 for a (C) main function with void return type
+        private static final int VOID_RET_VALUE = 214;
```
chrisseaton: Huh? Why?

mrigger: I figured out that the return value is actually whatever value is currently in `%eax`. I will update the change to return 0.

PR #299, opened 2016-06-08
Adding capability of Sulong to parse bitcode files

crbb:
Allow bitcode files to be decoded using su-run (using the same LLVM version as already used by Sulong) and assemble files in SulongTestSuites to pass the same tests as su-tests-sulong.


projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
```diff
@@ -95,6 +95,15 @@ public CallTarget parse(Source code, Node contextNode, String... argumentNames)
                     if (!context.isParseOnly()) {
                         parserResult.getStaticInits().call();
                     }
+                } else if (code.getMimeType().equals(LLVMLanguage.LLVM_BITCODE_MIME_TYPE)) {
```
mrigger: There is now a method `handleParserResult` to prevent code duplication.

mrigger:
Thanks a lot! Great to see that the parser passes (again) all the Sulong test cases!

chrisseaton:
I don't understand the comment 'su-run (using the same LLVM version as already used by Sulong)' - are we restricted to one LLVM version?

crbb:
The decoder currently uses the same version of LLVM as Sulong to maintain compatibility with Dragonegg. It has the ability to handle LLVM 3.8 but an option will be needed to switch between versions from the command line.

PR #300, opened 2016-06-10
Add additional truffle.h intrinsics

mickjordan:
Checking for a TruffleObject and creating byte[], cf truffle_read_string

### DIFF
--- a/include/truffle.h
+++ b/include/truffle.h
@@ -60,6 +60,7 @@ bool truffle_is_executable(void *object);
 bool truffle_is_null(void *object);
 bool truffle_has_size(void *object);
 bool truffle_is_boxed(void *object);
+bool truffle_is_truffle_object(void *object);
 
 // Execute:
 void *truffle_execute(void *object, ...);
@@ -126,6 +127,8 @@ void truffle_write_idx_b(void *object, int idx, bool value);
 
 // Strings
 void *truffle_read_string(const char *string);
+void *truffle_read_bytes(const char *bytes);
+void *truffle_read_n_bytes(const char *bytes, int n);
 
 #if defined(__cplusplus)
 }
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMTruffleReadBytes.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMTruffleReadBytes.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates.
+ *
+ * All rights reserved.
+ *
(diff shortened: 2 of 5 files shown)


include/truffle.h
```diff
@@ -126,6 +127,8 @@ void truffle_write_idx_b(void *object, int idx, bool value);
 
 // Strings
 void *truffle_read_string(const char *string);
+void *truffle_read_bytes(const char *bytes);
```
chrisseaton: This is in the string section of the API, but from the name alone it isn't clear it'll read until the first zero byte. Maybe `truffle_read_string_bytes`

mickjordan: Ah naming. Maybe read_cstring, because unlike read_string it does not create a Java String?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMTruffleReadBytes.java
```diff
@@ -0,0 +1,60 @@
...
+@NodeChild(type = LLVMAddressNode.class)
+public abstract class LLVMTruffleReadBytes extends LLVMAddressIntrinsic {
+    @Specialization
+    public Object executeIntrinsic(LLVMAddress value) {
+        byte c;
+        LLVMAddress adr = value;
+        int count = 0;
+        while ((c = LLVMMemory.getI8(adr)) != 0) {
+            count++;
+            adr = adr.increment(1);
```
mrigger: Could you use `Byte.Bytes` (for consistency with the other LLVM classes) here and at the other places where you iterate over an address space?

mrigger: Sorry for nitpicking.

mickjordan: I just followed the model that was used for truffle_read_string but am happy to use something else

mrigger: Thanks! I did not see that `truffle_read_string` used the unnamed constant directly. In `LLVMMemory` and other classes I used `Byte.Bytes`.

include/truffle.h
mickjordan: I think we may just need to document what these functions do. The "string" in truffle_read_string either identifies either the result type (Java String) or the input type (C string), but it reads until a NUL. In truffle_read_bytes we read until a NUL but create a Java byte[]. Both are really conversion style functions, not just "read" style functions.

chrisseaton: Ok, we can do that in a subsequent PR if you want to get this merged.

PR #301, opened 2016-06-13
Add TruffleObject eq/ne specializations to LLVMAddressCompareNode.

mickjordan:


mickjordan:
The error seems to be related to some download problem from linz. I don't see a way to retry the build without committing another change

chrisseaton:
I've restarted the build.

mrigger:
I updated the download link of Eclipse: #305.

mrigger:
Could you squash the commits before integration? In order to pass the git message gate check you have to rewrite the history anyway.

mickjordan:
On 6/19/16 2:37 AM, Manuel Rigger wrote:
>
> Could you squash the commits before integration? In order to pass the 
> git message gate check you have to rewrite the history anyway.
>
If you can tell me how to do that I will. I amended the commit with the 
bad message (why so anal by the way?)

Mick

mrigger:
This change was integrated with #308.

PR #309, opened 2016-06-20
Truffleobject compare

mickjordan:
Some related changes to prior PR

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/op/compare/LLVMAddressCompareNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/op/compare/LLVMAddressCompareNode.java
@@ -51,6 +51,22 @@ public abstract class LLVMAddressCompareNode extends LLVMI1Node {
             return val1 == val2;
         }
 
+        @SuppressWarnings("unused")
+        @Specialization(guards = "isCNull(val1)")
+        public boolean executeI1(LLVMAddress val1, TruffleObject val2) {
+            return false;
+        }
+
+        @SuppressWarnings("unused")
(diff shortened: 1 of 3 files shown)


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/op/compare/LLVMAddressCompareNode.java
```diff
@@ -51,6 +51,22 @@ public boolean executeI1(TruffleObject val1, TruffleObject val2) {
...
+        }
+
+        @SuppressWarnings("unused")
+        @Specialization(guards = "isCNull(val2)")
+
+        public boolean executeI1(TruffleObject val1, LLVMAddress val2) {
+            return false;
+        }
+
+        public static boolean isCNull(LLVMAddress val) {
```
mrigger: Do we need this guard? Can we not just return false in every case?

mickjordan: I originally just had it as false, but really the only comparison that should be legal is a comparison against NULL. Anything else is at best suspicious, so I thought it better to catch that case and deal with it if an when it arises.

chrisseaton: Does that answer your question @mrigger? Ready to merge?

mrigger: Yes, I am okay with it.

PR #311, opened 2016-06-28
Add pthread stub with pthread_create, pthread_join and pthread_exit

lxp:
pthread_create synchronously executes the start_routine function
pthread_exit is a no-op
pthread_join is a no-op

This stub allows running Argon2: https://github.com/P-H-C/phc-winner-argon2

### DIFF
--- a/mx.sulong/libs/pthread.h
+++ b/mx.sulong/libs/pthread.h
@@ -1,0 +1,15 @@
+#ifndef _PTHREAD_H
+#define _PTHREAD_H
+
+#ifndef NULL
+#define NULL 0
+#endif
+
+typedef long pthread_t;
+typedef void pthread_attr_t;
+
+extern int pthread_create (pthread_t *thread, const pthread_attr_t *attr, void *(*start_routine) (void *), void *arg);
+extern void pthread_exit (void *retval);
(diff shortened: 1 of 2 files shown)


mx.sulong/libs/pthread.h
```diff
@@ -0,0 +1,15 @@
+#ifndef _PTHREAD_H
```
mrigger: Do we need this header file? Can we not just include `<pthread.h>`?

lxp: I don't like it too, but we need the header file because `pthread_exit` usually has a `noreturn` attribute.
So, code compiled with the original header file won't run because it has an unreachable instruction after the `pthread_exit` call.

I wasn't sure where to put the header file, maybe we should move it to `include` instead of `mx.sulong/libs`?

mrigger: Okay, I understand. But then we need to use this header file instead of Clib's `pthread.h` header during compilation. Otherwise the unreachable instruction is still created.  To get this to work we have to modify the mx compile commands to include the header files. As an alternative we could also substitute the commands on the Java level and strip the unreachable instruction.

mrigger:
We agreed that it is easier to keep the substitutions in the application code for now.

PR #314, opened 2016-07-01
Provide intrinsics for abs, exp, ceil, floor, and rint

mrigger:
The cast functions have simple Java implementations (and no side effects).


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMCMathsIntrinsics.java
```diff
@@ -69,4 +73,94 @@ public double executeIntrinsic(double value) {
...
+        }
+
+    }
+
+    @NodeChild(type = LLVMDoubleNode.class)
+    public abstract static class LLVMFloor extends LLVMDoubleIntrinsic {
+
+        @Specialization
+        public double executeIntrinsic(double value) {
+            return Math.floor(value);
```
chrisseaton: `RFloor` and `Floor` both use `Math.floor` - is that right?

mrigger: You are right, I accidentally checked `Rfloor` in (at least the function is not registered in `LLVMRuntimeIntrinsicFactory`). I'll remove it again.

mrigger:
See #327.

PR #315, opened 2016-07-01
Update Truffle, mx, and graal-core versions

mrigger:
Replace methods deprecated in https://github.com/graalvm/truffle/commit/3f99fb6ba44b739109eddd3de388a56098390d4e.

### DIFF
--- a/projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
+++ b/projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
@@ -51,6 +51,7 @@ import com.oracle.truffle.api.Truffle;
 import com.oracle.truffle.api.TruffleLanguage.Env;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.nodes.RootNode;
+import com.oracle.truffle.api.source.MissingNameException;
 import com.oracle.truffle.api.source.Source;
 import com.oracle.truffle.api.vm.PolyglotEngine;
 import com.oracle.truffle.api.vm.PolyglotEngine.Builder;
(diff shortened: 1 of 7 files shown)


projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java
```diff
@@ -226,17 +227,21 @@ public static int executeMain(File file, Object... args) {
...
-        Source fromText = Source.fromText(codeString, "code string").withMimeType(LLVMLanguage.LLVM_IR_MIME_TYPE);
-        LLVMLogger.info("current code string: " + codeString);
-        return evaluateFromSource(fromText, args);
+        try {
+            Source fromText = Source.newBuilder(codeString).mimeType(LLVMLanguage.LLVM_IR_MIME_TYPE).build();
+            LLVMLogger.info("current code string: " + codeString);
+            return evaluateFromSource(fromText, args);
+        } catch (MissingNameException e) {
+            throw new AssertionError(e);
+        }
```
woess: you should provide a name, otherwise this will always throw, won't it?

mrigger: Thanks! I did not realize that this method was not used (or just by a single test case that is not part of the CI). I will remove the method altogether.

woess: If it's not tested, it doesn't work. :wink:

mrigger: Yes, I think we need to use coverage tools to identify untested code.

PR #325, opened 2016-07-05
Use the newly available JVMCI JDK 8 binaries in Travis

mrigger:
This change frees us from the burden of having to build JVMCI ourselves.


mrigger:
We cannot download it since it requires a logged in user.

PR #330, opened 2016-07-10
Restore JRuby's cext tests

chrisseaton:


chrisseaton:
This now tests building and running a proper off-the-shelf C extension, as well as some of Ruby's C extension API tests.

Issue #323, opened 2016-07-04
Standalone su link tool


chrisseaton:
We need a tool for linking su files that isn't part of mx, so it can be shipped as a command with GraalVM. Should probably be implemented in Java with a shell script launcher.

PR #331, opened 2016-07-10
Java link tool

chrisseaton:
Move the link tool into Java so that we can make it usable from GraalVM builds.


chrisseaton:
Fixes #323

mrigger:
I found some code style violations and confirmed them with `mx checkstyle` and `mx pylint`.

I will check what is wrong with the gate configuration.

PR #336, opened 2016-07-14
Implement TCK test cases

pekd:
Currently there are no TCK test cases. This change adds a very basic TCK test suit (with many test cases being disabled) as well as the first change towards successfully running it. Typecasts for calls from foreign languages are implemented.


projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
```diff
@@ -0,0 +1,415 @@
...
+import com.oracle.truffle.llvm.test.LLVMPaths;
+import com.oracle.truffle.llvm.tools.Clang;
+import com.oracle.truffle.llvm.tools.Clang.ClangOptions;
+import com.oracle.truffle.llvm.tools.Opt;
+import com.oracle.truffle.llvm.tools.Opt.OptOptions;
+import com.oracle.truffle.llvm.tools.Opt.OptOptions.Pass;
+import com.oracle.truffle.tck.TruffleTCK;
+
+/**
+ * This is the way to verify your language implementation is compatible.
```
mrigger: Compatible with what? I think you can just remove this comment.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
```diff
@@ -0,0 +1,415 @@
...
+ *
+ */
+public class LLVMTckTest extends TruffleTCK {
+    private static final String PATH = LLVMPaths.LOCAL_TESTS + "/../interoptests";
```
mrigger: Can you add the `PATH` constant in [`LLVMPaths`](https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/LLVMPaths.java)?

pekd: The same constant is also defined in [`LLVMInteropTest`](https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMInteropTest.java). Should I change it there also?

mrigger: Yes, please!

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    @Override
+    protected String invalidCode() {
+        // @formatter:off
+        return
+            "f unction main() {\n" +
+            "  retu rn 42;\n" +
+            "}\n";
+        // @formatter:on
```
chrisseaton: Shouldn't this be invalid IR, rather than invalid C?

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
```diff
@@ -0,0 +1,410 @@
...
+    protected String multiplyCode(String firstName, String secondName) {
+        // @formatter:off
+        /*
+        return
+            "function multiply(" + firstName + ", " + secondName + ") {\n" +
+            "  return " + firstName + " * " + secondName + ";\n" +
+            "}\n";
+        // @formatter:on
+         */
+        return null;
```
chrisseaton: How come this is null?

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
```diff
@@ -0,0 +1,410 @@
...
+    public void timeOutTest() throws Exception {
+        super.timeOutTest();
+    }
+
+    @Override
+    @Ignore
+    @Test
+    public void testCoExistanceOfMultipleLanguageInstances() throws Exception {
+        super.testCoExistanceOfMultipleLanguageInstances();
+    }
```
chrisseaton: Are there individual reasons why these tests are ignored? Should we document that in comments?

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
pekd: But anything which is not valid IR is invalid IR, including valid/invalid C?

chrisseaton: True, but if that was the intention the TCK wouldn't ask you to provide your own code. It could just use some junk for anything. I think the idea was more that you'd provide something that checked that your parser was really detecting errors in your language.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
pekd: Because there is some constraint in the eval-code, and therefore this test case does not work right now, and I also did not yet figure out how to fix it.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
pekd: Yes, there are. Either they check something with structs, or they check something with function passing, which is not yet implemented.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/interop/LLVMTckTest.java
pekd: Maybe, but then one single test case is not sufficient. I can specify "invalid" IR which results in parse errors, I can specify "invalid" IR which parses without errors but produces runtime errors when the parsed code is executed, and I can specify something which is not IR at all, like I did in the beginning. And I think all of those cases should be checked. But I also think that this has more to do with the parser itself, than with interop …

chrisseaton:
Is this ready to go?

pekd:
It should be ready to go, if you do not see any problems anymore.

chrisseaton:
You just need to update the branch and then I'll merge.

PR #340, opened 2016-07-17
Support Truffle objects in globals

chrisseaton:


chrisseaton:
This works for Ruby code storing managed Ruby objects in global variables.

chrisseaton:
This is ready for initial review. Ignore the JRuby test failure for now.

Key comment:

https://github.com/graalvm/sulong/pull/340/files#diff-a39e5fc5437b3cb9251c567960163ba6R36

Problem I need to solve:

https://github.com/graalvm/sulong/pull/340/files#diff-ace6de25373f780bf4543873d50a681fR1051

mickjordan:
This will be useful for the 30 odd global variables in the R FFI, which I have been ignoring so far (or mapping to functions assuming they are read-only)

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/base/LLVMNode.java
```diff
@@ -48,4 +50,28 @@
 
     public abstract void executeVoid(VirtualFrame frame);
 
+    protected boolean isLLVMAddress(Object object) {
+        return object instanceof LLVMAddress;
+    }
+
+    @SuppressWarnings("unused")
+    protected boolean isUninitialized(VirtualFrame frame, LLVMGlobalVariableStorage globalVariableStorage) {
```
mrigger: I'm not sure if it is a good idea to have `isUninitialized`, `isInitialized`, `isNative` and `isManaged` in the `LLVMNode` class, since the methods are not really related to it. I think one of the two options would work better:
1. Directly call the guard, which would create a direct correspondence between the guard (e.g., `globalVariableStorage.isInitialized`) and initialize method (e.g., `globalVariableStorage.initialize`).
2. Make a helper class for it, if you think that it is likely that we will replace the guard test in the future.

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMGlobalVariableStorage.java
```diff
@@ -0,0 +1,130 @@
...
+     * MANAGED are final states. When you have assigned explicitly a native or managed value that
+     * will be the state of the global variable storage for the rest of the program.
+     * 
+     * UNINITIALIZED -> INITIALIZED -> (NATIVE | MANAGED)
+     */
+
+    private enum State {
+        UNINITIALIZED,
+        INITIALIZED,
+        NATIVE,
```
mrigger: Maybe we should rename `NATIVE` to `NATIVELY_INITIALIZED`?

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
```diff
@@ -1048,6 +1048,18 @@ private LLVMExpressionNode visitConstant(EObject type, Constant constant) throws
...
+                /*
+                 * Global variables in C can be 'extern', but still managed because they're defined
+                 * in another managed file. We should be able to detect that by looking for existing
+                 * global variables with the same name, so I would think that we should just need to
+                 * add
+                 * 
+                 * && !globalVars.containsKey(globalVariable)
+                 * 
+                 * to the condition below, but if I do that things go pretty badly wrong and we end
+                 * up crashing.
```
mrigger: I think the check does not work since a native `extern` variable can also be defined in the same file:

```
extern int a;
int a = 4;

int main() {
	return a;
}
```

mrigger:
Looks good! I noted a few things in inline comments.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/base/LLVMNode.java
chrisseaton: Fixed by putting them in a separate file and using `@ImportStatic`.

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMGlobalVariableStorage.java
chrisseaton: Fixed.

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java
chrisseaton: Is the original code wrong anyway then? In your example it will try to look up `a` using NFI, but if the variable is defined in LLVM IR (in the same file or not) it won't be found like that will it?

PR #344, opened 2016-07-20
Add truffle_read_n_string and fix truffle_read_n_bytes interop intrinsic

lxp:
`truffle_read_n_string` is necessary for Ruby's `rb_str_new` function that allows specifying a length.

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleReadNString.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleReadNString.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are
+ * permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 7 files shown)


lxp:
I accidentally pushed something using my private mail address that is not registered with graalvmbot. I updated my pull request to only include my JKU address.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/interop/LLVMTruffleReadNString.java
```diff
@@ -0,0 +1,55 @@
...
+    @Specialization
+    public Object executeIntrinsic(LLVMAddress value, int n) {
+        LLVMAddress adr = value;
+        int count = n < 0 ? 0 : n;
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < count; i++) {
+            sb.append((char) Byte.toUnsignedInt(LLVMMemory.getI8(adr)));
+            adr = adr.increment(Byte.BYTES);
+        }
+        return sb.toString();
```
chrisseaton: I'm afraid I don't think we can use `StringBuilder` in the fast path like this. You either need a boundary, or to use a manual `byte[]` and then `new String()` (with a boundary).

lxp: Thank you for pointing that out. I have added a TruffleBoundary.

PR #348, opened 2016-07-26
Implement interop for arrays

pekd:
This change adds the possibility of using llvm arrays in foreign languages as well as using foreign language arrays in llvm code (with a limitation: JavaInterop cannot be used for llvm arrays, because it cannot handle arrays without a size).


mrigger:
This change adds flatted types for pointers. There are already cases where I implemented such types for functions and vectors. I think it is not a good idea to follow this road, since it blows up many switches of which most do not care which kind of type the object pointed to has.

Can we instead have a `LLVMType` that has a reference to a `LLVMBaseType` and a parent `LLVMType`? Then an int pointer would be represented as a `LLVMType` that has an `ADDRESS` as a    `LLVMBaseType` (which the existing switches can use) and a `LLVMType` as parent that has `INT` as its `LLVMBaseType`.

pekd:
I changed it in the following way: the `LLVMBaseType` has a single `ADDRESS` once again, but this type is encapsulated in an `LLVMType`, along with a `pointee` type if it is a pointer. With this change, a lot of switches get simpler. However, I left the `LLVMRuntimeType` with the flattened types, because there are very few locations (right now: 4) in the code where a switch statement is used on a `LLVMRuntimeType`.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LLVMType.java
```diff
@@ -0,0 +1,48 @@
...
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.parser;
+
+public class LLVMType {
+    public final LLVMBaseType type;
```
mrigger: Could we have getters and `private final` modifiers for the `type` and `pointee`? Then we would not have to make the fields part of the API of this class. Otherwise, this change looks good to me.

PR #352, opened 2016-07-30
Basic source sections and error reporting

chrisseaton:
When running a Ruby extension which has something that Sulong can't support, or a legitimate bug in your C code, we previously saw this.

```
RubyTruffleError: internal implementation error - ClassCastException com.oracle.truffle.object.basic.DynamicObjectBasic cannot be cast to java.lang.Long com.oracle.truffle.llvm.nodes.impl.func.LLVMCallUnboxNode$LLVMI64CallUnboxNode.executeI64(LLVMCallUnboxNode.java:104)
from jruby/lib/ruby/truffle/truffle/openssl.rb:10:in `<main>'
from jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:55:in `gem_original_require'
from jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:55:in `require'
from ../test.rb:1:in `<main>'

```

With this pull request applied we see this.

```
jruby/lib/ruby/truffle/truffle/openssl.rb:10:in `require_relative': internal implementation error - LLVM error in assignment of %2 in the first basic block in function @Init_openssl in jruby/lib/ruby/truffle/openssl/openssl.su@f49022b58d1b25cb6ccd08ca395d4ec584f3cd40_ossl.ll (RubyTruffleError)
from jruby/lib/ruby/truffle/truffle/openssl.rb:10:in `<main>'
from jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:55:in `gem_original_require'
from jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:55:in `require'
from ../test.rb:1:in `<main>'
```

This is enough information to pin the problem down to a single line. If the error isn't in an assignment, currently you'll see just the basic block information. I will look at more specific information for non-assignments later (it's tricky because how do you uniquely name it?)


mrigger:
I tested the change for performance regressions since I was not sure whether the execute method in `LLVMBasicBlockNode` would result in additional deoptimization points and thus more code. I could not verify that yet, since the change results in additional PE errors on the `oggenc` benchmark. We first have to fix these PE errors in Graal before merging the change.

mickjordan:
Thanks Chris, this will be very helpful to me

mrigger:
I tested again and found that probably my build was broken, which caused the PE errors. There does not seem to be performance regressions.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMBasicBlockNode.java
```diff
@@ -55,23 +60,43 @@
...
     }
 
     @ExplodeLoop
     public int executeGetSuccessorIndex(VirtualFrame frame) {
         for (LLVMNode statement : statements) {
-            statement.executeVoid(frame);
+            try {
+                statement.executeVoid(frame);
+            } catch (ControlFlowException e) {
+                controlFlowExceptionProfile.enter();
```
mrigger: The branch profiling is not absolutely necessary here: Only `abort`, `exit`, and the LLVM IR `unreachable` are `ControlFlowException`s that are mostly not executed. Since they are mostly not executed, they are also not compiled due to the branch profiling in the [basic block node](https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/others/LLVMBlockNode.java#L83). I still think that the profiling is useful here, since we might want to add other control flow excecptions in the future.

PR #356, opened 2016-08-01
Add exp and pow intrinsics for the C standard library functions

mrigger:
Some measurements on microbenchmarks that call the functions with different values in a loop:

pow:
    native call: 585
    intrinsification before Intel PR: 2007
    intrinsification after Intel PR: 162

exp:
    native call: 160
    intrinsification before Intel PR: 160
    intrinsification after Intel PR: 56


mrigger:
See https://github.com/graalvm/graal-core/pull/198

PR #342, opened 2016-07-20
Support multi-line inline assembly

PrinzKatharina:

PR #360, opened 2016-08-03
Support multi-line inline assembly V2

PrinzKatharina:


mrigger:
Integrated with #342.

PR #361, opened 2016-08-03
Add Argon2 tests

lxp:


mx.sulong/mx_sulong.py
```diff
@@ -435,7 +454,7 @@ def runLLVM(args=None):
             libNames.append(arg)
         else:
             sulongArgs.append(arg)
-    return mx.run_java(getCommonOptions(libNames) + vmArgs + getClasspathOptions() + ['-XX:-UseJVMCIClassLoader', "com.oracle.truffle.llvm.LLVM"] + sulongArgs, jdk=mx.get_jdk(tag='jvmci'))
+    return mx.run_java(getCommonOptions(libNames) + vmArgs + getClasspathOptions() + ['-XX:-UseJVMCIClassLoader', "com.oracle.truffle.llvm.LLVM"] + sulongArgs, out=out, jdk=mx.get_jdk(tag='jvmci'))
```
mrigger: Why do we need `out=out`?

lxp: To support redirection of stdout.
Argon2 has tests, where stdout must be compared to a reference output (see `runTestArgon2Kats()`).

PR #367, opened 2016-08-05
Implement interop for foreign function pointers

pekd:
The existing interop for functions allows to call LLVM functions from a foreign function.
This change implements the opposite direction: a foreign function can be passed to a LLVM function which expects a function pointer. The LLVM function can then call this function, which executes the foreign function.


mrigger:
Would it be possible to also provide a test case (with moderate effort)?

mrigger:
If there are already test cases that cover the important cases then it's okay. I just wondered since this change did not enable more test cases, and I am sure to break everything without test cases. ;)

chrisseaton:
If you update then I'll merge.

Issue #363, opened 2016-08-03
Fix argument conversion for indirect native calls


lxp:
LLVMAddress should be converted to long during an indirect native call.

```c
#include <string.h>

int main(void) {
        char test[512];
        static void *(*const volatile memset_v)(void *, int, size_t) = &memset;
        memset_v(test, 0, 512);
        return 0;
}
```

```
$ mx su-clang -S -emit-llvm -o memset-test.ll memset-test.c
$ mx su-run -ea -esa memset-test.ll
Exception in thread "main" java.lang.AssertionError: java.io.IOException: java.lang.AssertionError: memset[long, int, long] expected arg 0 to be java.lang.Long, not com.oracle.truffle.llvm.types.LLVMAddress
	at com.oracle.truffle.llvm.LLVM.evaluateFromSource(LLVM.java:274)
	at com.oracle.truffle.llvm.LLVM.executeMain(LLVM.java:249)
	at com.oracle.truffle.llvm.LLVM.main(LLVM.java:203)
Caused by: java.io.IOException: java.lang.AssertionError: memset[long, int, long] expected arg 0 to be java.lang.Long, not com.oracle.truffle.llvm.types.LLVMAddress
	at com.oracle.truffle.api.TruffleLanguage$LanguageImpl.eval(TruffleLanguage.java:585)
	at com.oracle.truffle.api.vm.PolyglotEngine.evalImpl(PolyglotEngine.java:565)
	at com.oracle.truffle.api.vm.PolyglotEngine.eval(PolyglotEngine.java:532)
	at com.oracle.truffle.api.vm.PolyglotEngine.eval(PolyglotEngine.java:469)
	at com.oracle.truffle.llvm.LLVM.evaluateFromSource(LLVM.java:271)
	... 2 more
Caused by: java.lang.AssertionError: memset[long, int, long] expected arg 0 to be java.lang.Long, not com.oracle.truffle.llvm.types.LLVMAddress
	at com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionHandle.checkArgs(HotSpotNativeFunctionHandle.java:119)
	at com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionHandle.call(HotSpotNativeFunctionHandle.java:85)
	at com.oracle.truffle.llvm.nodes.impl.func.LLVMCallNode$LLVMFunctionCallChain$1.execute(LLVMCallNode.java:295)
[...]

PR #369, opened 2016-08-08
Fix the LLVMAddress arg conversion in indirect calls with native functions

mrigger:
Fixes https://github.com/graalvm/sulong/issues/363.


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMCallNode.java
```diff
@@ -292,6 +293,11 @@ public Object execute(VirtualFrame frame) {
                         Object[] arguments = frame.getArguments();
```
chrisseaton: Should we be profiling the number of arguments and exploding the loop?

mrigger: Thanks for noticing the missing `@ExplodeLoop`! I added an assert for the case where the length of the `arguments` array is not a compilation constant. I will add test cases for indirect function calls of native functions with variadic arguments in a subsequent PR, where such cases could potentially happen.

PR #371, opened 2016-08-09
Add lifetime analysis tests

mrigger:
This change adds lifetime test cases and auto-generated reference output. Having lifetime test cases prevents performance regressions, especially when we switch to the binary LLVM IR parser.


mrigger:
@chrisseaton do you think that including the generated reference outputs in the repository is a good idea? I thought it would be better to include them than putting them on lafo, since it might be useful to track the changes in the test files if the lifetime analysis changes.

chrisseaton:
It is quite a lot of files. Do you ever imagine them changing? Is this a one-time thing?

mrigger:
They could change, but probably not more than once or twice. Maybe I upload them on lafo?

chrisseaton:
I think that might be best.

PR #383, opened 2016-08-17
Fix some issues in the bitcode parser

jakre:


projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/Constants.java
```diff
@@ -31,6 +31,9 @@
 
 public class Constants {
 
+    public static final boolean USE_BINARY_PARSER = false;
```
mrigger: Could we have that as a `LLVMBaseOption`? Then you and others can test the binary parser without having to modify the source code.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/LLVMTestSuite.java
```diff
@@ -53,7 +53,7 @@
 @RunWith(Parameterized.class)
 public class LLVMTestSuite extends RemoteTestSuiteBase {
 
-    private static final int TEST_TIMEOUT_TIME = 15000;
+    private static final int TEST_TIMEOUT_TIME = -1;
```
mrigger: Is that change accidental?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/base/LLVMContext.java
```diff
@@ -49,6 +49,7 @@
 
     private final List<RootCallTarget> staticInitializers = new ArrayList<>();
     private final List<RootCallTarget> staticDestructors = new ArrayList<>();
+    private final List<RootCallTarget> constructorFunctions = new ArrayList<>();
```
mrigger: I think the `constructorFunctions` here contains the same variables as the `staticInitializers` in `LLVMContext`, `LLVMGlobalRootNode` and `LLVMVisitor` (see [`LLVMVisitor in line 415`](https://github.com/graalvm/sulong/blob/6c78c9b8afad7a5fb1440959c43812af4ff40ab3/projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/LLVMVisitor.java#L415)). The name `staticInitializers` is misleading and should probably be called `staticConstructors` so that it is obvious that they relate to the `staticDestructors`. I think for this commit it is sufficient to keep the changes in the binary bitcode parser classes.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMGlobalRootNode.java
```diff
@@ -104,13 +104,16 @@ protected Object executeIteration(VirtualFrame frame, int iteration, Object[] ar
         if (printExecutionTime) {
             startExecutionTime = System.currentTimeMillis();
         }
+
+        executeConstructorFunctions();
```
mrigger: We execute the constructor functions the first time in the [LLVM class](https://github.com/graalvm/sulong/blob/6c78c9b8afad7a5fb1440959c43812af4ff40ab3/projects/com.oracle.truffle.llvm/src/com/oracle/truffle/llvm/LLVM.java#L189) since they also need to be executed when a function is called over interop where this node is not constructed. I think we should not change this logic in this node.

mrigger:
Thanks for the PR. The code looks good and clean! We can merge the PR once you fixed the issues I noted in the comments.

PR #389, opened 2016-08-22
Add an Haiku architecture description to the documentation

mrigger:


ARCHITECTURE.md
```diff
@@ -0,0 +1,76 @@
+# Architectural Haiku
+
```
chrisseaton: Maybe add a link to what a Haiku is? In case people think you've just written a poem.

PR #400, opened 2016-08-28
Allow passing the null-function to native

chrisseaton:
Also remove unsupported reasons more clearly, which is how I found this problem.


mrigger:
The changes look good. I would be good to have a test case for this. I only came up with `qsort` when thinking about C functions that expect function pointers. However, no null pointer can be passed to the function. A minimal test case could thus look something like this:
```
#include <stdlib.h>

int values[] = { 1, 2, 3, 4, 5 };

int main() {
  int (*f)(const void *, const void *) = NULL;
  qsort(values, 0, sizeof(int), f);
}
```

Still, it would be better to have a new test suite that compiles a predefined shared library to have better coverage for this test case.

chrisseaton:
Yes I'm working on a few managed/native interactions so it would be useful to have a place to write tests for these with a compiled native library. I'll have to talk to you to figure out how to do it.

PR #404, opened 2016-08-28
Remove the global state in LLVMTypeHelper

chrisseaton:
`LLVMTypeHelper` has global state that is set to the IR file which happened to be parsed last. If you only parse bitcode files I'm not sure it's ever set.

We really need to start eliminating these static variables.

### DIFF
--- a/projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/LLVMBitcodeVisitor.java
+++ b/projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/LLVMBitcodeVisitor.java
@@ -216,7 +216,8 @@ public class LLVMBitcodeVisitor implements ModelVisitor {
                         store = LLVMMemI32CopyFactory.create(globalVarAddress, (LLVMAddressNode) constant, new LLVMI32LiteralNode(size), new LLVMI32LiteralNode(0), new LLVMI1LiteralNode(false));
                     } else {
                         Type t = global.getValue().getType();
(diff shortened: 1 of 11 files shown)


projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/LLVMBitcodeVisitor.java
```diff
@@ -353,4 +357,8 @@ public void visit(FunctionDefinition method) {
     @Override
     public void visit(Type type) {
     }
+
+    public LLVMTypeHelper getTypeHelper() {
```
mrigger: Could we expose this method in `LLVMParserRuntime`? Then we could leave the node factories stateless.

chrisseaton: I'm not sure what you mean. This method is so that `LLVMBitcodeFunctionVisitor` can get the type helper. Neither of those classes has an `LLVMParserRuntime`.

And the node factories can't be stateless unless you pass in a type helper for each call that needs it.

mrigger: The `LLVMVisitor` implements `LLVMParserRuntime` to provide some common functionality to the factories. The `LLVMParserRuntime` provides some context information such as the return stack slot, stack pointer slot, bit alignment computation and so on. By using the `LLVMParserRuntime`, we do not have to extend the factories to have more and more parameters. I do not know about the `LLVMBitcodeFunctionVisitor` though.

chrisseaton: Fixed.

Issue #403, opened 2016-08-28
Fix cross-module managed global variables


chrisseaton:
At the moment each IR file creates a new `LLVMGlobalVariableStorage` for each global variable it knows about. This breaks the system for storing managed objects in global variables, as the storage and specialisation information for that lives in `LLVMGlobalVariableStorage`. We need to share `LLVMGlobalVariableStorage` objects globally.

chrisseaton:
I found the `CorrectExternalVariableLinking` option - why is this `false` by default @mrigger? Setting it to true doesn't fix this issue though.

mrigger:
The option should fix the issue that the Graal NFI is used to resolve a global variable, even when it is defined in another source file (see #395). I think the cross-module `LLVMGlobalVariableStorage` sharing could build on this change.

The option makes start-up slower, since it iterates over the global variables of the bitcode libraries and the source to be parsed. Since support for it is currently rarely needed, I disabled it by default until we have finished the integration of the binary parser.

chrisseaton:
I'm sure it can be simple than this. Could we make it lazy? When you see a global you create an `LLVMGlobalVariableStorage`, which does no native lookup up front. If you see the place where the global is defined (as opposed to declared) you mark the global storage object. When all your files are loaded and you start to run the program and first use a global variable, it could then look for the native symbol if it hasn't been marked as defined.

mrigger:
Yes, that would be possible and more efficient. I decided for the other implementation since it less complex.

chrisseaton:
We already solve this problem for functions don't we? I think we should have a global variable registry, like we have a function registry. I'll do a PR to suggest that.

chrisseaton:
Fixed by #417

PR #417, opened 2016-09-03
New global variable registry and lazy resolution system

### LINKED ISSUE #380: Fix linking of extern variables
Currently, linking of extern symbols is broken when the external symbol is defined in one of the LLVM IR files (see #340). The current lookup logic (see `LLVMVisitor`) is as following:
```
if ("external".equals(linkage)) {
    long getNativeSymbol = nativeLookup.getNativeHandle(globalVarName);
    LLVMAddress nativeSymbolAddress = LLVMAddress.fromLong(getNativeSymbol);
    return factoryFacade.createLiteral(nativeSymbolAddress, LLVMBaseType.ADDRESS);
} else {
    Object findOrAllocateGlobal = findOrAllocateGlobal(globalVariable);
    assert findOrAllocateGlobal != null;
    return factoryFacade.createLiteral(findOrAllocateGlobal, LLVMBaseType.ADDRESS);
}
```
Per default, external symbols are looked up by using the Graal NFI while they should first be looked up in the bitcode files. 

It is not enough to add a condition that checks if there is already an entry for a global variable (`&& !globalVars.containsKey(globalVariable)`) since space for all global variables is initially allocated, even if the symbol is looked up later and the allocated global variable space not used. Even if the redundant allocation would be fixed inside the `LLVMVisitor` it would not be enough to find global extern variables that are defined in other files.

To fix this issue, we have to traverse all bitcode files, determine the unresolved extern variables and resolve them (with the Graal NFI). We could either keep patch addresses and patch them after constructing the ASTs (with one traversal) or first determine the unresolved global variables, and then construct the right references in the visitor (with two traversals).

chrisseaton:
Fixes #380 and #403.

Makes resolution (figuring out if it's native or managed, and looking up an external native symbol) lazy.


projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/NodeFactoryFacadeImpl.java
```diff
@@ -409,19 +411,42 @@ public LLVMFunctionDescriptor createFunctionDescriptor(String name, LLVMRuntimeT
...
+        boolean isStatic = "internal".equals(linkage) || "private".equals(linkage);
+        boolean isExtern = "external".equals(linkage);
+
+        String name = globalVariable.getName();
+
+        NativeResolver nativeResolver = new NativeResolver() {
+
+            @Override
+            public LLVMAddress resolve() {
+                return LLVMAddress.fromLong(runtime.getNativeHandle(name));
```
chrisseaton: The package that `LLVMGlobalVariableDescriptor` is in means that it doesn't have access to things like `LLVMParserRuntime`, so I wasn't sure how to let it resolve the external global variable apart from with this closure.

chrisseaton:
@mrigger the one failure is in the GCC C torture tests and is because the variable `but` is being declared more than once. It does appear in multiple test files - are you loading all the test files into the same Sulong context? I made it an assertion failure to have variables declared more than once, as it's a linker error. What do you think?

projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/NodeFactoryFacadeImpl.java
mrigger: I think that having the closure is okay. If we have to introduce more such abstractions we can consider refactoring the packages.

mrigger:
The `TestGCCSuite` that we use for these test cases uses `LLVM.executeMain` for every test case. Thus, there should be a new `LLVMContext` for every test case. It could be that there is still some static field that we have not eliminated yet, which stores the context.

Otherwise, it could be that a bitcode library (see `-Dsulong.DynamicBitcodeLibraries`) defines a variable with non-`static` linkage that conflicts with another declaration.

I will try to reproduce the issue locally and see if I can find out the problem.

chrisseaton:
You could store the file in which the variable was declared in `LLVMGlobalVariableDescriptor`, and then when you see it declared a second time print the current file and the file of the first declaration.

PR #421, opened 2016-09-04
Allow boolean objects to be converted to an integer

chrisseaton:
Arrises when we have a Java `Boolean`, which we are treating as a Ruby `VALUE` (`void *`). This is converted to an integer in a Ruby extension. I think it makes sense to allow this Java type be converted to a raw pointer and then integer value, even though we don't want to do it with arbitrary managed objects.

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/cast/LLVMToI8Node.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/cast/LLVMToI8Node.java
@@ -132,6 +132,11 @@ public abstract class LLVMToI8Node extends LLVMI8Node {
     public abstract static class LLVMAddressToI8Node extends LLVMToI8Node {
 
         @Specialization
+        public byte executeI8(boolean from) {
+            return (byte) (from ? 1 : 0);
+        }
+
+        @Specialization
         public byte executeI8(LLVMAddress from) {
             return (byte) from.getVal();
         }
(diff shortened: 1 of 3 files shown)


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/cast/LLVMToI8Node.java
```diff
@@ -132,6 +132,11 @@ public byte executeI8(LLVM80BitFloat from) {
     public abstract static class LLVMAddressToI8Node extends LLVMToI8Node {
```
mrigger: Would it make sense to change the type of `fromNode` to `LLVMExpressionNode`? I think it is misleading to suggest that the node's input can always return an address, which I think, is, what using the node class should imply. I saw that  we still use the `LLVMAddressNode` at similar locations too.

mrigger:
Could you add a test case for that? This change could potentially break at some point, and we should not only rely on the Ruby gate for such cases.

mrigger:
Maybe this would be a better solution: We can have single `LLVMTo*Node`s with specializations for each source type. Then we would not have to special case single types.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/cast/LLVMToI8Node.java
chrisseaton: I thought that `LLVMAddressNode` included Truffle objects? And `Boolean` is a bit like that - a managed object.

mrigger: It is because each node that produces a value is an `LLVMExpressionNode` which has an execute method that returns an `Object`. This method exists so that function calls can easily execute the children to get a generic value for each of them.

I still think that it is confusing that an address to I8 (and other integer types) cast node can receive a boolean input value (which we could make transparent by having the `LLVMTo*CastNode`) and that, if we want to integrate the change, should have a test case for it.

If this issue blocks you we can merge the change and handle it properly the next time we stumble on such a case.

chrisseaton: Ah I filled in the I8 and I32 ones by default, but I only need the I64 one. I'll remove the others.

I'm also trying the `LLVMTo*Node` idea now.

PR #425, opened 2016-09-05
Add installation documentation into "docs" folder

pointhi:
General documentation was copied from README.md including a little update of the labsjdk version

But more important, I added distribution specific installation instructions (for archlinux)


docs/INSTALLATION.md
```diff
@@ -0,0 +1,78 @@
+## Installation of Sulong
```
mrigger: I think we can remove the paragraph under this headline, since the `README.md` already covers the basic installation procedure.

docs/index.md
```diff
@@ -7,3 +7,4 @@
 * [Truffle language interoperability and Sulong](INTEROP.md)
 * [Sulong-related publications](PUBLICATIONS.md)
 * [Sulong's test cases](TESTS.md)
+* [How to get Sulong installed](INSTALLATION.md)
```
mrigger: Can we change that to `[Distribution-specific installation of the external dependencies](DEPENDENCIES.md)` or so? I think this file name and description match its content better.

mrigger:
LGTM!

chrisseaton:
It's just the OCA check then.

PR #427, opened 2016-09-05
Support Forward References in the Binary Parser

jakre:


mrigger:
+1 for the static factory methods!

PR #430, opened 2016-09-06
Fix default numeric formatting category when executing mx

pointhi:
Some test cases have string comparisons which require usage of the english number formatting.


mrigger:
Great to have this fixed!

PR #435, opened 2016-09-06
Avoid immediate exiting on LLVM version mismatch while executing 'mx su-tests'

SwapnilGaikwad:
- LLVM version check ensures version > 3.x for LLVM binaries while executing e.g. 'mx su-tests'. On failure to find version number (from the output of  '<program> --version' command), script shows the same output to user and exit. Instead, when 'None' is returned from the function, script downloads and uses those LLVM program executables.


mx.sulong/mx_sulong.py
```diff
@@ -755,7 +755,7 @@ def getLLVMVersion(llvmProgram):
     versionString = getVersion(llvmProgram)
     printLLVMVersion = re.search(r'(clang |LLVM )?(version )?(3\.\d)', versionString, re.IGNORECASE)
     if printLLVMVersion is None:
-        exit("could not find the LLVM version string in " + str(versionString))
+        exit("Could not find the LLVM version 3.x using command '" + llvmProgram + " --version'.\nYou may also use the LLVM binaries from " + _toolDir +"tools/llvm/bin" )
```
mrigger: I think we should not print that the user can use the downloaded LLVM binaries at this place. This function should only try to extract the version from the LLVM program that it receives. If the passed `llvmProgram` is `None` or specifies a non-existing path, then we have to fix this at the caller of the function.

README.md
```diff
@@ -65,11 +65,11 @@ Next, you need to download a recent
 [labsjdk](http://www.oracle.com/technetwork/oracle-labs/program-languages/downloads/index.html).
 Extract it inside the `sulong-dev` directory:
 
-    tar -zxf labsjdk-8u92-jvmci-0.17-linux-amd64.tar.gz
+    tar -zxf labsjdk-8u92-jvmci-0.20-linux-amd64.tar.gz
```
mrigger: There is already an open PR for fixing the JVMCI version in the documentation: #425

mrigger:
LGTM! Thanks for the PR! We can merge it as soon as your name appears on the [OCA list](http://www.oracle.com/technetwork/community/oca-486395.html).

SwapnilGaikwad:
Closing PR as #444 incorporates these changes.

PR #442, opened 2016-09-08
Add implementation of toString() to some classes

pointhi:
to simplify debugging of those datastructures


mrigger:
LGTM! You could also have used `String.format` for the `toString` methods with many String concatenations, to make them more readable.

Issue #402, opened 2016-08-28
Support converting managed objects to native


chrisseaton:
This pattern is common in Ruby:

```c
native_function(RSTRING_PTR(string));
```

Where `string` is a Ruby object, and `RSTRING_PTR` is a function that exposes the `char*` in it. We normally support this by returning another managed object that just provides the illusion of a `char*` but actually reads and writes the managed Ruby string.

This doesn't work when we want to pass the string to native.

I think an option here would be for Sulong to give Ruby the chance to convert the string to a native representation when it escapes to native. We could have a new interop message `NATIVE`, and it would work something like this:

Sulong sees a managed object being passed to a native function. It would pass it the `NATIVE` message. For the Ruby string object, this message would be implemented by calling `Unsafe.allocateMemory` and copying the managed string data into native memory, and then returning the pointer. Ruby already has multiple internal representations for strings, and this new native representation would just be another one of those. The Ruby program could keep running using the native string as normal. Ruby returns the address of this native memory from the `NATIVE` message, and Sulong passes that to the native function.

grimmerm:
In general, I like the idea.
What I like most is that we can do that as late as possible, i.e., at the very last moment before the string escapes to native code. Some kind of worst case scenario. 

However, I think NATIVE message should not become part of the Truffle Interop API, this is too Sulong/Ruby specific. It is possible to create additional and specialized messages though.

chrisseaton:
Yes we could do it via an `INVOKE` with a special name, and Sulong could know to send the message based on looking at the `ObjectType` to see it's a Ruby object.

chrisseaton:
Having solved some other issues, this is now the thing that is blocking me and we can see where it needs to go.

```
Caused by: com.oracle.truffle.api.dsl.UnsupportedSpecializationException: Unexpected values provided for LLVMAddressToI64NodeGen@1c4c5e0: [org.jruby.truffle.interop.cext.CExtString@6ffd08ac], [CExtString]
	at com.oracle.truffle.api.dsl.internal.SpecializationNode.unsupported(SpecializationNode.java:481)
	at com.oracle.truffle.api.dsl.internal.SpecializationNode.uninitialized(SpecializationNode.java:396)
	at com.oracle.truffle.llvm.nodes.impl.cast.LLVMToI64NodeFactory$LLVMAddressToI64NodeGen$UninitializedNode_.executeLong_(LLVMToI64NodeFactory.java:709)
[...]

PR #446, opened 2016-09-09
Converting objects to native as they escape

chrisseaton:
This is the basic implementation for #402, specialised just for the Ruby use case, which is at https://github.com/jruby/jruby/commit/c30e0ecc0b47f8027d350a3d7a4f7e0961e188c7.

We probably want to do this with something like the adapter that @mrigger already suggested, but I wanted to show the code that does enough to work.

I decided to repurpose the `UNBOX` message.

This lets us do:

```
$ JRUBY_TRUFFLE_NATIVE_OPENSSL=true GRAAL_HOME=$SULONG_HOME jt run --graal -e "require 'openssl'; p OpenSSL::Random.random_bytes(12).bytes"
[200, 158, 154, 243, 111, 122, 42, 200, 18, 94, 89, 124]
```

Where the `openssl` gem is using the native system `libssl.so` to actually generate the random bytes, and is writing into a raw Ruby `char*` string (it thinks), which Ruby then continues to use.

@mrigger @grimmerm @mickjordan

### DIFF
--- a/docs/INTEROP.md
+++ b/docs/INTEROP.md
@@ -45,6 +45,11 @@ TODO
 
 `object()` sends `EXECUTE`
 
+A reference to a foreign object being converted to an integer (possibly to make
+a native call) sends `UNBOX` and treats the returned value as a native memory
+address. This behaviour is likely to change in the future, but currently
+enables foreign objects to convert to a native representation if possible.
+
 ## Intrinsic functions
 
 ### `string.h`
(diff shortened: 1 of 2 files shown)


chrisseaton:
So what do we want to do about making this work for both FastR and JRuby? Maybe this explicit class check is fine and we'll do it for classes from both languages? Maybe we need a new interop message? Is there some other nicer way we can check to see if `UNBOX` will do what we want? What do you think you need @mickjordan and what would you want in the Sulong code base @grimmerm?

mrigger:
I think that we should not have any code in Sulong that is specific to another Truffle implementation, so I vote against the explicit class check.

chrisseaton:
I've modified to just send `UNBOX` for now. We can discuss some better options for the longer term elsewhere.

PR #448, opened 2016-09-13
Add LLVM Cauldron presentation to publications list

chrisseaton:


mrigger:
Integrated with #473.

PR #449, opened 2016-09-13
Add gcc-bc test suite for BitCode parser

SwapnilGaikwad:
- Adding a 'gcc-bc' test suite for BitCode parser, equivalent to the 'gcc' test suite for .ll files.
- Actual code change is quite little, most of the files conating list of files to be tested.
- Separate test suite required causing duplication of files listing names of files to be executed/tested because not all the files running with .ll parser runs with BitCode parser.
- Adding a new mx target 'su-travis4' to run tests on BitCode parser. This addition would increase the overall regression tests time.
- Generation of .bc files from FORTRAN files is done in 2 stages (.f90 -> .ll -> .bc) as a quick fix. Please suggest if it can be more elegant.

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -67,6 +67,7 @@ are remote test cases.
 | test suite | type   | config | description                              |
 |------------|--------|--------|------------------------------------------|
 | llvm       | remote | yes    | LLVM's test suite                        |
+| llvm-bc    | remote | yes    | LLVM's test suite for BitCode parser     |
 | interop    | local  | no     | Truffle interoperability test suite      |
 | polyglot   | local  | no     | Minimal Polyglot engine test             |
 | tck        | local  | no     | Test suite to certify Truffle compliance |
@@ -75,6 +76,7 @@ are remote test cases.
 | asm        | local  | no     | Inline assembler tests                   |
 | sulong     | local  | no     | Sulong's own primary test suite          |
 | gcc        | local  | yes    | GCC's test suite                         |
+| gcc-bc     | local  | yes    | GCC's test suite for BitCode parser      |
 | main-arg   | local  | no     | Tests arguments to the main function     |
 | bench      | local  | no     | Language Benchmark game benchmarks       |
 | compile    | local  | yes    | GCC's compile-only test cases            |


mrigger:
Thanks, the changes look good! I especially like the good commit messages, that each commit contains a logical change, and the detailed PR description.

I believe that by executing both the working GCC and LLVM test cases with the bitcode parser, we have sufficient regression tests until we completely replace the parser for the textual format with the binary parser. After replacing the textual parser, we will execute all test cases with the binary parser anyway. If we add more partial test suites for the bitcode parser, we would end up having a gate that takes almost twice as long to execute as the current one.Thus, I believe we should stop integrating more bitcode parser test suites for now.

projects/com.oracle.truffle.llvm.tools/src/com/oracle/truffle/llvm/tools/GCC.java
```diff
@@ -60,9 +61,23 @@ public static void compileToLLVMIR(File toBeCompiled, File destinationFile) {
...
+            // Creating .ll file for the FORTRAN source file
+            String llFileGenerationcommand = tool + " -I " + LLVMBaseOptionFacade.getProjectRoot() + "/../include -S " + dragonEggOption() + " -fplugin-arg-dragonegg-emit-ir  -o /tmp/temp.ll " +
+                            toBeCompiled.getAbsolutePath();
+            ProcessUtil.executeNativeCommandZeroReturn(llFileGenerationcommand);
+
+            // Converting .ll file to .bc file
+            String llToBCConversionCommand = LLVM_AS_PATH.toString() + " /tmp/temp.ll -o " + destinationFile;
+            ProcessUtil.executeNativeCommandZeroReturn(llToBCConversionCommand);
+        } else {
+            String[] command = new String[]{tool, "-I " + LLVMBaseOptionFacade.getProjectRoot() + "/../include", "-S", dragonEggOption(), "-fplugin-arg-dragonegg-emit-ir", "-o " + destinationFile,
```
mrigger: Could you factor out the logic for creating the bitcode file that is used both in the `if` and `else` branch?

projects/com.oracle.truffle.llvm.tools/src/com/oracle/truffle/llvm/tools/GCC.java
```diff
@@ -60,9 +61,23 @@ public static void compileToLLVMIR(File toBeCompiled, File destinationFile) {
...
+        if (destinationFile.getName().endsWith(".bc")) {
+            // TODO: Check for more elegant way to generate .bc file for FORTRAN source file
+
+            // Creating .ll file for the FORTRAN source file
+            String llFileGenerationcommand = tool + " -I " + LLVMBaseOptionFacade.getProjectRoot() + "/../include -S " + dragonEggOption() + " -fplugin-arg-dragonegg-emit-ir  -o /tmp/temp.ll " +
+                            toBeCompiled.getAbsolutePath();
+            ProcessUtil.executeNativeCommandZeroReturn(llFileGenerationcommand);
+
+            // Converting .ll file to .bc file
+            String llToBCConversionCommand = LLVM_AS_PATH.toString() + " /tmp/temp.ll -o " + destinationFile;
```
mrigger: Could you add the functionality for using `llvm-as` in the `LLVMAssembler` class, and remove it from here? Also, you can reference the `llvm-as` tool path by looking it up from the `LLVMToolPaths` class.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/GCCBCTestSuite.java
```diff
@@ -0,0 +1,99 @@
...
+            if (!undefinedReturnCode) {
+                Assert.assertEquals(byteCodeFile.getAbsolutePath(), expectedResult, truffleResult);
+            }
+        } catch (Throwable e) {
+            recordError(tuple, e);
+            throw e;
+        }
+    }
+
+    private static int truncate(int retValue) {
```
mrigger: Can we add this logic in `TestSuiteBase`? Otherwise, we have the same code in both GCC test suite classes.

projects/com.oracle.truffle.llvm.test/suites-configs/gccbc/gcc.c-torture/execute/working.exclude
```diff
@@ -0,0 +1,84 @@
+gcc.c-torture/execute/20010518-2.c
```
mrigger: Why is `working.exclude`?

projects/com.oracle.truffle.llvm.test/suites-configs/gccbc/gcc.dg/unsorted/working.exclude
```diff
@@ -0,0 +1,28 @@
+gcc.dg/pubtypes-3.c
```
mrigger: Also here is a a `working.exclude` file.

projects/com.oracle.truffle.llvm.test/suites-configs/gccbc/gcc.c-torture/execute/working.exclude
mrigger: @SwapnilGaikwad told me offline that these files can be processed by the `.ll` parser, but not by the `.bc` parser.

SwapnilGaikwad:
Update code to ingest above comments and excluded the test cases causing the crash during regression tests.

PR #452, opened 2016-09-14
Consider the specified TargetDataLayout when computing Type Size and Alignment in the Binary Parser

jakre:


projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/util/DataLayoutConverter.java
```diff
@@ -0,0 +1,114 @@
...
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.parser.bc.impl.util;
+
+import com.oracle.truffle.llvm.parser.LLVMBaseType;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class DataLayoutConverter {
```
mrigger: Could we have a new project `com.oracle.truffle.llvm.parser.base` or so to which we move classes that are both used by the binary and textual parser? I think that makes maintenance and debugging easier, since we already know that these classes are "supposed to work" due to their usage in the textual parser. This applies also to some other classes in this PR.

projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/util/DataLayoutType.java
```diff
@@ -0,0 +1,41 @@
...
+
+public enum DataLayoutType {
+    INTEGER,
+    FLOAT,
+    VECTOR,
+    POINTER,
+    AGGREGATE,
+    STACK,
+    STACK_OBJECT,
+    INTEGER_WIDTHS
```
mrigger: What is `INTEGER_WIDTHS`? If it is the type for variable width integers (integers != `INTEGER`) then maybe we could name it `VARIABLE_WIDTH_INTEGER` or so?

projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/util/LLVMParserAsserts.java
```diff
@@ -0,0 +1,62 @@
...
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.parser.bc.impl.util;
+
+import java.util.Arrays;
+
+public final class LLVMParserAsserts {
```
mrigger: I think that this class could also be moved to the `parser.base` package

projects/com.oracle.truffle.llvm.parser.base/src/com/oracle/truffle/llvm/parser/base/datalayout/DataLayoutParser.java
```diff
@@ -117,4 +114,13 @@ private static DataLayoutType getDataType(String string) {
         }
     }
 
+    private static Object[] assertNoNullElement(Object[] objects) {
```
mrigger: It's not really an issue for a small helper method but this method is a duplicate of the one in `LLVMParserAsserts`.

jakre: I deleted the corresponding class.

projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/util/DataLayoutConverter.java
jakre: Done.

projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/util/DataLayoutType.java
jakre: This was taken from an existing class in the Textparser.

PR #453, opened 2016-09-15
Remove hard-coded temporary file in the bitcode GCC tests

mrigger:


projects/com.oracle.truffle.llvm.tools/src/com/oracle/truffle/llvm/tools/GCC.java
```diff
@@ -62,10 +63,15 @@ public static void compileToLLVMIR(File toBeCompiled, File destinationFile) {
...
+        File interimFile;
+        try {
+            interimFile = File.createTempFile("interim", ".ll");
+        } catch (IOException e) {
+            throw new AssertionError(e);
+        }
 
         if (destinationFile.getName().endsWith(".bc")) {
-            destinationLLFileName = interimLLFileName;
+            destinationLLFileName = interimFile.getName();
```
SwapnilGaikwad: Maybe 'interimFile.getAbsolutePath()' shall do the trick and also it will create files in temp dir as we are using this name in .ll file generation command.

mrigger: Thanks!

PR #455, opened 2016-09-16
Add new mx commands to simplify local verification of the code base

pointhi:


mx.sulong/mx_sulong.py
```diff
@@ -1155,6 +1175,8 @@ def checkNoHttp(args=None):
     'su-travis-sulong' : [travisTestSulong, ''],
     'su-travis-jruby' : [travisJRuby, ''],
     'su-travis-argon2' : [travisArgon2, ''],
+    'su-ecj' : [runEcjTestCases, ''],
+    'su-checkcode' : [checkCode, ''],
```
mrigger: Can we rename the command to `su-basic-checkcode` or so? Otherwise the command name suggests that it executes all our linter tools.

mx.sulong/mx_sulong.py
```diff
@@ -509,6 +504,31 @@ def runTests(args=None):
         command = testCases[testSuiteName]
         command(vmArgs)
 
+def checkCode(args=None):
```
mrigger: One (optional) fancy thing to do would be to use a table like in `mx su-tests` to let the user select a subset of static analysis checks to be executed.

mx.sulong/mx_sulong.py
```diff
@@ -1155,6 +1175,8 @@ def checkNoHttp(args=None):
     'su-travis-sulong' : [travisTestSulong, ''],
     'su-travis-jruby' : [travisJRuby, ''],
     'su-travis-argon2' : [travisArgon2, ''],
+    'su-ecj' : [runEcjTestCases, ''],
```
mrigger: I think we do not need a separate mx command for that. If you are using `javac` per default you can use `mx build --jdt <path>` to compile with the Eclipse compiler and if you are using the Eclipse compiler per default (by having set the `JDT` environment variable) you can use `mx build --force-javac`.

PR #456, opened 2016-09-16
Some additional debug flags

pointhi:
* ```-Dsulong.Verbose``` enable extensive debug output of metadata parser
* ```-Dsulong.LLVM``` specify used version of LLVM parser


projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/options/LLVMBaseOptionFacade.java
```diff
@@ -38,6 +38,10 @@ public static boolean debugEnabled() {
         return LLVMOptions.getParsedProperty(LLVMBaseOption.DEBUG);
     }
 
+    public static boolean verboseEnabled() {
+        return LLVMOptions.getParsedProperty(LLVMBaseOption.VERBOSE);
```
mrigger: Maybe use `debugEnabled() && LLVMOptions.getParsedProperty(LLVMBaseOption.VERBOSE);` here? Then one has to use `-Dsulong.Debug=true -Dsulong.Verbose=true` to get verbose debugging. If you do not want that, I would rename the option to `-Dsulong.VerboseDebug`.

PR #460, opened 2016-09-18
Use a guard instead of ClassCastException for specializing in the address arg node

mrigger:


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/func/LLVMArgNode.java
```diff
@@ -186,11 +186,18 @@ public LLVM80BitFloat execute80BitFloat(VirtualFrame frame) {
 
         public abstract int getIndex();
 
-        @Specialization(rewriteOn = ClassCastException.class)
+        @Specialization(guards = "isAddress(frame)")
         public Object executePointee(VirtualFrame frame) {
+            /*
+             * We copy the address to prevent its immediate allocation if the address escapes.
```
eregon: What is `the address immediate allocation`?

mrigger: I will replace the comment with the one from the original commit message (https://github.com/graalvm/sulong/commit/9e8b274672c1132814116735f6fd122818f36c39): "Copying the address objects prevents that otherwise virtualized objects that are used in a phi together with a materialized passed address object also have to be materialized."

PR #466, opened 2016-09-21
Some Metadata Parser improvements

pointhi:
* remove/implement some TODO's
  * now we support unicode strings in Nodes like ```MetadataString```
  * improve handling of References to the Symbol Table inside MetadataNodes
* fix a little metadata parser issue
* add typechecks for curently not used arguments (to detect scheme problems more easily)
* Extend ```DwTagRecord``` and add ```DwLanguageRecord```


jakre:
Looks good to me!

PR #469, opened 2016-09-22
Add a document to describe Safe Sulong and its memory error findings

mrigger:
Thanks to @eregon for verifying the error, analyzing its cause, and recommending a fix!

### DIFF
--- a/docs/SAFE-SULONG.md
+++ b/docs/SAFE-SULONG.md
@@ -1,0 +1,219 @@
+Safe Sulong is a version of Sulong that only uses managed memory (Java allocations)
+to prevent temporal and spatial memory safety errors. You can read about
+its architecture in
+[Sulong: Memory Safe and Efficient Execution of LLVM-Based Languages](http://ssw.jku.at/General/Staff/ManuelRigger/ECOOP16-DS.pdf).
+While also other tools (such as Clang's AddressSanitizer) can find memory
+errors, Sulong strives to provide memory safety "per default", since not
+all developers use static analysis tools.
+
+Safe Sulong is currently in development and not available on Github. The version
+on Github relies on unmanaged memory with which memory safety violations
+can still occur.
+
+# The Computer Language Benchmarks Game: fasta-redux C gcc #2 program
+
+The Computer Language Benchmarks Game "shootouts" are small benchmarks
+to compare the performance of different languages. The
+[fasta-redux C gcc #2 program](https://benchmarksgame.alioth.debian.org/u64q/program.php?test=fastaredux&lang=gcc&id=2)
+benchmark contains a buffer overflow that we detected with Safe Sulong:
+
+```c
+/* The Computer Language Benchmarks Game
+ * http://benchmarksgame.alioth.debian.org/
+
+   based on C++ program by Andrew Moon
+   contributed by Damien Grassart
+*/
+
+#define _GNU_SOURCE
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+#define MIN(x, y) ((x < y) ? x : y)
+#define LINELEN 60
+#define SLOTS 4095
+
+struct aminoacid {
+   char c;
+   float p;
+};
+
+static struct aminoacid *lu[SLOTS + 1];
+
+static void repeat_fasta(const char *alu, size_t n)
+{
+   const size_t alulen = strlen(alu);
+   char buf[alulen + LINELEN];
+   size_t pos = 0, bytes;
+
+   memcpy(buf, alu, alulen);
+   memcpy(buf + alulen, alu, LINELEN);
+   while (n) {
+      bytes = MIN(LINELEN, n);
+      fwrite_unlocked(buf + pos, bytes, 1, stdout);
+      putchar_unlocked('\n');
+      pos += bytes;
+      if (pos > alulen)
+         pos -= alulen;
+      n -= bytes;
+   }
+}
+
+static void acc_probs(struct aminoacid *table)
+{
+   struct aminoacid *iter = table;
+
+   while ((++iter)->c) {
+      iter->p += (iter-1)->p;
+   }
+   for (int i = 0; i <= SLOTS; ++i) {
+      while (i > (table->p * SLOTS))
+         ++table;
+      lu[i] = table;
+   }
+}
+
+static float rng(float max)
+{
+   const unsigned int IM = 139968, IA = 3877, IC = 29573;
+   static unsigned int seed = 42;
+
+   seed = (seed * IA + IC) % IM;
+   return max * seed / IM;
+}
+
+static char nextc()
+{
+   float r;
+   struct aminoacid *iter;
+
+   r = rng(1.0f);
+   iter = lu[(int)(r * SLOTS)];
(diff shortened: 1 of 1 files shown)


docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
+Safe Sulong is a version of Sulong that only uses managed memory (Java allocations)
+to prevent temporal and spatial memory safety errors. You can read about
+its architecture in
+[Sulong: Memory Safe and Efficient Execution of LLVM-Based Languages](http://ssw.jku.at/General/Staff/ManuelRigger/ECOOP16-DS.pdf).
+While also other tools (such as Clang's AddressSanitizer) can find memory
+errors, Sulong strives to provide memory safety "per default", since not
+all developers use static analysis tools, while not sacrificing peak performance.
```
eregon: Hard to read, maybe mixing two things here: always|per default and strive to not sacrificing peak perf. I think splitting in two sentences  would help.

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
...
+can still occur.
+
+# The Computer Language Benchmarks Game: fasta-redux C gcc #2 program
+
+The Computer Language Benchmarks Game "shootouts" are small benchmarks
+to compare the performance of different languages. The
+[fasta-redux C gcc #2 program](https://benchmarksgame.alioth.debian.org/u64q/program.php?test=fastaredux&lang=gcc&id=2)
+benchmark contains a buffer overflow that we detected with Safe Sulong:
+
+```
```
eregon: "```c"

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
...
+   fputs_unlocked(">THREE Homo sapiens frequency\n", stdout);
+   acc_probs(homosapiens);
+   random_fasta(homosapiens, n * 5);
+   putchar_unlocked('\n');
+
+   return 0;
+}
+```
+
+In the `acc_probs` function the first loop computes accumulates probabilities:
```
eregon: accumulated but even better cumulative probabilities

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
...
+
+```
+   while ((++iter)->c) {
+      iter->p += (iter-1)->p;
+   }
+```
+
+The second last element should have an accumulated probability of `1`.
+For example, when `iub` is passed to the function (with 16 elements),
+then the function expects the last element `(table + 15)->p` to be `1`.
```
eregon: the second last element probability `table[14].p` to be `1`.

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
...
+Due to rounding errors, the last element is actually `0.999999762`.
```
eregon: , the computed cumulative probability is ...

docs/SAFE-SULONG.md
mrigger: I removed the peak performance aspect, since we do not have numbers yet to provide it.

docs/SAFE-SULONG.md
mrigger: I also fixed this in the other code snippets.

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
...
+
+When `i=SLOTS (4095)` the condition should be false, since `4095 > (1 * 4095)`.
+However, since the last element has the probability `0.999999762` the condition
+is `4095 > (0.999999762 * 4095)` and hence `false`. The loop does not terminate
+and continues to read `table->p` where `table` points to out-of-bounds memory.
+The program does not crash when executed by the (Native) Sulong contained
+in this repo, and when executed as an executable compiled with static
+compilers. It does not crash since it reads mapped memory, so the error
+goes unnoticed. The function terminates after reading a negative "garbage
+value". Safe Sulong detectes the out-of-bounds access through its bounds checks
```
eregon: No, it finishes after reading a positive value >= 1.0 (in my case 3.07825471e+22).

docs/SAFE-SULONG.md
mrigger: Ok, I also used cumulative instead accumulated at the other occurences.

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
...
+and terminates the program.
+
+## Fix
+
+A fix for the buffer overflow is to make the probability check more robust
+by taking into account that floating point computations have rounding errors:
+
+For example, the condition `i > (table->p * SLOTS)` can be replaced by
+`fabs(i - table->p * SLOTS) < 0.00001` to prevent the out-of-bounds read.
```
eregon: Still fairly dangerous if the last probability was more away(above or below) than  0.00001 from 1.0

docs/SAFE-SULONG.md
mrigger: I also changed `(table + 15)->p` to `table[15].p`.

docs/SAFE-SULONG.md
mrigger: Done.

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,218 @@
...
+```
+   for (int i = 0; i <= SLOTS; ++i) {
+      while (i > (table->p * SLOTS))
+         ++table;
+      lu[i] = table;
+   }
+```
+
+When `i=SLOTS (4095)` the condition should be false, since `4095 > (1 * 4095)`.
+However, since the last element has the probability `0.999999762` the condition
```
eregon: check for all occurrences of `last` ;)
Or maybe say you don't consider the null real-last element

docs/SAFE-SULONG.md
mrigger: I change the sentence to `The function terminates after reading a "garbage
value" that lets the condition yield false.` since the value is not guaranteed to be stable anyway.

docs/SAFE-SULONG.md
mrigger: I think with four digits after the comma we are safe, for this program at least.

docs/SAFE-SULONG.md
mrigger: Done.

mrigger:
Thanks for your feedback! I tried to address all issues and updated the PR.

docs/SAFE-SULONG.md
```diff
@@ -0,0 +1,219 @@
...
+   for (int i = 0; i <= SLOTS; ++i) {
+      while (i > (table->p * SLOTS))
+         ++table;
+      lu[i] = table;
+   }
+```
+
+When `i=SLOTS (4095)` the condition should be false, since `4095 > (1 * 4095)`.
+However, since the second last element has the probability `0.999999762` the condition
+is `4095 > (0.999999762 * 4095)` and hence `false`. The loop does not terminate
```
eregon: `true` !

mrigger: Ah, right.

PR #478, opened 2016-09-26
Implement basic Signal handling (closes #277)

### LINKED ISSUE #277: Add support for Signal handlers
We currently [ignore signal handler registrations](https://github.com/graalvm/sulong/blob/master/mx.sulong/libs/signal.c). It think it would not be too hard to handle them on the Java side with `sun.misc.SignalHandler`. For example, you can do the following in Java:

```
SignalHandler sh = new SignalHandler() {
            @Override
            public void handle(Signal sig) {
                System.out.println(sig.getName() + " " + sig.getNumber());
            }
        };
        Signal.handle(new Signal("HUP"), sh);
        Signal.handle(new Signal("INT"), sh);
        Signal.handle(new Signal("TERM"), sh);
        while (true) {
            try {
                Thread.sleep(500);
            } catch (InterruptedException e) {
                e.printStackTrace();
            }
        }
```
`Signal.handle` registers the Signal handler. So when you, e.g., use `kill -INT <process id>` in Linux, then instead of terminating, the `handle` method is called. In Sulong we can intrinsify the C `signal` function and remember the function to then call it (with the signal number as an argument) when a signal occurs.

pointhi:
now we can use signals inside sulong (with a few exceptions for signals which are used by Java itself)

There is a big TODO regarding stack size calculation, but the rest should be fine.

This PR fixes #277 

There exists a second (much more complex) signal handling system, which is not supported by this patch: [sigaction](https://en.wikipedia.org/wiki/Sigaction)

Here is a simple test program:

```c
#include <stdio.h>
#include <signal.h>
#include <unistd.h>
#include <stdlib.h>
#include <string.h>

void sig_handler(int signo) {
    printf(" * received SIGNAL: %d (%s)\n", signo, strsignal(signo));
}

int main(void)
{
    if (signal(SIGINT, sig_handler) == SIG_ERR) {
        printf("\ncan't catch SIGINT\n");
        exit(1);
    }

    // A long wait so that we can easily issue a signal to this process
    int i = 1;
    while(i) {
        sleep(1);
        printf(".\n");
    }
    return 0;
}
```

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
@@ -32,6 +32,9 @@ package com.oracle.truffle.llvm.nodes.impl.intrinsics.c;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import com.oracle.truffle.api.Truffle;
 import com.oracle.truffle.api.CompilerDirectives;
@@ -45,6 +48,7 @@ import com.oracle.truffle.api.frame.FrameDescriptor;
 import com.oracle.truffle.api.source.SourceSection;
 import com.oracle.truffle.llvm.nodes.base.LLVMExpressionNode;
 import com.oracle.truffle.llvm.nodes.base.LLVMNode;
+import com.oracle.truffle.llvm.nodes.base.LLVMThreadNode;
 import com.oracle.truffle.llvm.nodes.impl.base.LLVMContext;
 import com.oracle.truffle.llvm.nodes.impl.base.LLVMFunctionNode;
 import com.oracle.truffle.llvm.nodes.impl.base.LLVMLanguage;
@@ -117,13 +121,19 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
     private static final long TMP_SIGNAL_STACK_SIZE_KB = 512;
     private static final long TMP_SIGNAL_STACK_SIZE_BYTE = TMP_SIGNAL_STACK_SIZE_KB * 1024;
 
-    private static final class LLVMSignalHandler implements SignalHandler {
+    private static final class LLVMSignalHandler implements SignalHandler, LLVMThreadNode {
 
+        private final Signal signal;
         private final LLVMFunctionDescriptor function;
+        private final LLVMContext context;
         private RootCallTarget callTarget;
         private final LLVMStack stack = new LLVMStack();
 
+        Lock lock = new ReentrantLock();
+        private AtomicBoolean isRunning = new AtomicBoolean(true);
+
         private LLVMSignalHandler(Signal signal, LLVMFunctionDescriptor function) throws IllegalArgumentException {
+            this.signal = signal;
             this.function = function;
             LLVMFunctionNode functionNode = LLVMFunctionLiteralNodeGen.create(function);
 
@@ -131,7 +141,7 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
             LLVMI32LiteralNode sigNumArg = new LLVMI32LiteralNode(signal.getNumber());
             LLVMExpressionNode[] args = {signalStack, sigNumArg};
 
-            LLVMContext context = LLVMLanguage.INSTANCE.findContext0(LLVMLanguage.INSTANCE.createFindContextNode0());
+            context = LLVMLanguage.INSTANCE.findContext0(LLVMLanguage.INSTANCE.createFindContextNode0());
 
             LLVMUnresolvedCallNode callNode = new LLVMUnresolvedCallNode(functionNode, args, LLVMRuntimeType.VOID, context);
 
@@ -142,6 +152,8 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
                                             SourceSection.createUnavailable("", null),
                                             new FrameDescriptor(), ""));
 
+            context.registerThread(this);
+
             if (function.equals(LLVM_SIG_DFL)) {
                 Signal.handle(signal, SignalHandler.SIG_DFL);
             } else if (function.equals(LLVM_SIG_IGN)) {
@@ -149,6 +161,15 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
             } else {
                 Signal.handle(signal, this);
             }
+        }
+
+        @Override
+        protected void finalize() throws Throwable {
+            super.finalize();
+            stop();
+            // TODO: wait for finish of running handle?
+            context.unregisterThread(this);
+            stack.free();
         }
 
         @Override
@@ -162,13 +183,39 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
              * @note I made such a program, and I had to delete the handler manually, so this
              * sentence is probably not true
              */
-            callTarget.call();
+
+            lock.lock();
+            try {
+                if (isRunning.get()) {
+                    callTarget.call();
+                }
+            } finally {
+                lock.unlock();
+            }
         }
 
         public LLVMFunctionDescriptor getFunction() {
             return function;
         }
 
+        @Override
+        public void stop() {
+            isRunning.set(false);
+        }
+
+        @Override
+        public void awaitFinish() {
+            stop();
+
+            // wait until handle is finished
+            lock.lock();
+            lock.unlock();
+        }
+
+        @Override
+        public String toString() {
+            return "LLVMSignalHandler [signal=" + signal + ", lock=" + lock + ", isRunning=" + isRunning + "]";
+        }
     }
 
     /**
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
@@ -49,7 +49,7 @@ import com.oracle.truffle.api.frame.FrameDescriptor;
 import com.oracle.truffle.api.source.SourceSection;
 import com.oracle.truffle.llvm.nodes.base.LLVMExpressionNode;
 import com.oracle.truffle.llvm.nodes.base.LLVMNode;
-import com.oracle.truffle.llvm.nodes.base.LLVMThreadNode;
+import com.oracle.truffle.llvm.nodes.base.LLVMThread;
 import com.oracle.truffle.llvm.nodes.impl.base.LLVMContext;
 import com.oracle.truffle.llvm.nodes.impl.base.LLVMFunctionNode;
 import com.oracle.truffle.llvm.nodes.impl.base.LLVMLanguage;
@@ -100,6 +100,7 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
 
     private static final Map<Integer, LLVMSignalHandler> registeredSignals = new HashMap<>();
 
+    @TruffleBoundary
     public static int getNumberOfRegisteredSignals() {
         return registeredSignals.size();
     }
@@ -138,8 +139,8 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
     }
 
     // TODO: stack handling should work without predefined sizes,...
-    private static final long TMP_SIGNAL_STACK_SIZE_KB = 512;
-    private static final long TMP_SIGNAL_STACK_SIZE_BYTE = TMP_SIGNAL_STACK_SIZE_KB * 1024;
+    private static final long SIGNAL_STACK_SIZE_KB = 512;
+    private static final long SIGNAL_STACK_SIZE_BYTE = SIGNAL_STACK_SIZE_KB * 1024;
 
     /**
      * Registers a signal handler using sun.misc.SignalHandler. Unfortunately, using signals in java
@@ -154,7 +155,7 @@ public abstract class LLVMSignal extends LLVMFunctionNode {
      * wait until the signal was handled (which is not guaranteed because of the asynchronous
      * behavior in our implementation).
      */
-    private static final class LLVMSignalHandler implements SignalHandler, LLVMThreadNode {
+    private static final class LLVMSignalHandler implements SignalHandler, LLVMThread {
 
         private final Signal signal;
         private final LLVMFunctionDescriptor function;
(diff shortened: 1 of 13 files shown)


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+     * Handling conversation from Signal number to Signal Object in a clean way.
+     *
+     * documentation: http://man7.org/linux/man-pages/man7/signal.7.html
+     */
+    private enum Signals {
+
+        // POSIX.1-1990
+        SIG_HUP("HUP"),
```
mrigger: Can you assign these constants an explicit number? For example, `SIG_HUP("HUP", 1)`. It's error-prone to rely on the order of enum constants.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,245 @@
...
+        SIG_CLD("CLD"),
+        SIG_PWR("PWR"),
+        SIG_INFO("INFO"),
+        SIG_LOST("LOST"),
+        SIG_WINCH("WINCH"),
+        SIG_UNUSED("UNUSED");
+
+        public static Signals decode(int code) {
+            for (Signals currentSignal : values()) {
+                if (currentSignal.signal() != null && currentSignal.signal().getNumber() == code) {
```
mrigger: `currentSignal.signal() != null` is always `true`, isn't it?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,245 @@
...
+                    return currentSignal;
+                }
+            }
+            return null;
```
mrigger: Maybe we can throw an `AssertionError` here?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        private final Signal signal;
+
+        Signals(String signalName) {
+            Signal constructedSignal;
+            try {
+                constructedSignal = new Signal(signalName);
+            } catch (IllegalArgumentException e) {
+                // it seems like this signal is not available on your system
```
mrigger: Can we print an error/debug message here with `LLVMLogger`?

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMStack.java
```diff
@@ -76,6 +76,10 @@ public void free() {
         isFreed = true;
     }
 
+    public static long allocate(long size) {
```
mrigger: For integration of this PR we need to have an own stack for signal handlers. I guess we also have to lock on it once we enter a signal handler to avoid race conditions.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,245 @@
...
+import com.oracle.truffle.llvm.types.memory.LLVMStack;
+
+import sun.misc.Signal;
+import sun.misc.SignalHandler;
+
+@GenerateNodeFactory
+@NodeChildren({@NodeChild(type = LLVMI32Node.class, value = "signal"), @NodeChild(type = LLVMFunctionNode.class, value = "handler")})
+public abstract class LLVMSignal extends LLVMFunctionNode {
+
+    private static final LLVMFunctionDescriptor LLVM_SIG_DFL = LLVMFunctionDescriptor.create(0);
```
mrigger: Can you add a comment on how these macro definitions look like on your system? It will help us later, if they differ on other systems.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,245 @@
...
+    public LLVMFunctionDescriptor executeAddress(int signal, LLVMFunctionDescriptor handler) {
+        return setSignalHandler(signal, handler);
+    }
+
+    @TruffleBoundary
+    private static LLVMFunctionDescriptor setSignalHandler(int signalId, LLVMFunctionDescriptor function) {
+        Signals decodedSignal = Signals.decode(signalId);
+
+        if (decodedSignal == null) {
+            LLVMLogger.error("unknown signal: " + signalId);
```
mrigger: We can add a `CompilerDirectives.transferToInterpreterAndInvalidate();` after the null check and remove the @TruffleBoundary to make the code "safe for compilation".

projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/memory/LLVMStack.java
mrigger: Actually, since signal handler support does not need to be fast (yet), we can also allocate the memory once we enter the signal handler and deallocate it again once we leave it.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
pointhi: I don't rely on the sequence of the signals. To be more specific, only a few signals even have a specified signal number defined in the POSIX standard (https://en.wikipedia.org/wiki/Unix_signal#POSIX_signals)

Because of that, the signal number, as well as which signals are supported on the system is evaluated at runtime in this enum.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
pointhi: it's false if the signal is not supported on the system (which can happen). Look onto the constructor for more details

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
pointhi: Probably with ```LLVMLogger.info``` You have to note there are always some signals which a system doesn't support, so we have to evaluate this at runtime

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
pointhi: done

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
pointhi: done

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
mrigger: Okay, thanks for the clarification!

projects/com.oracle.truffle.llvm.test/tests/c/signal/signal003.c
```diff
@@ -0,0 +1,49 @@
+#include <signal.h>
```
mrigger: Great that you added thorough test cases!

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,425 @@
...
+            lock.lock();
+            try {
+                if (function.equals(LLVM_SIG_DFL)) {
+                    Signal.handle(signal, SignalHandler.SIG_DFL);
+                    stack.free();
+                } else if (function.equals(LLVM_SIG_IGN)) {
+                    Signal.handle(signal, SignalHandler.SIG_IGN);
+                    stack.free();
+                } else {
+                    Signal.handle(signal, this);
```
mrigger: Is it possible that you just allocate the stack here, if you free it immediately in the other cases?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,425 @@
...
+            }
+        }
+
+        public boolean isRunning() {
+            return isRunning.get();
+        }
+
+        @Override
+        @TruffleBoundary
+        protected void finalize() throws Throwable {
```
mrigger: Finalizers are not guaranteed to be executed. Does your approach still work if they aren't?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,425 @@
...
+        } catch (IllegalArgumentException e) {
+            LLVMLogger.error("could not register signal with id " + signalId + " (" + signal + ")");
+            return LLVM_SIG_ERR;
+        }
+
+        return returnFunction;
+    }
+
+    // TODO: stack handling should work without predefined sizes,...
+    private static final long TMP_SIGNAL_STACK_SIZE_KB = 512;
```
mrigger: I think we can just rename this to `SIGNAL_STACK_SIZE_KB` since we already have the TODO here. I do not think that this will ever change anyway.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,425 @@
...
+
+        @Override
+        public void handle(Signal arg0) {
+            try {
+                if (!globalSignalHandlerLock.tryLock(HANDLE_MAX_WAITING_TIME, TimeUnit.MILLISECONDS)) {
+                    LLVMLogger.error("could not execute signal handler. Sulong can currently only execute one signal at once!");
+                    return;
+                }
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
mrigger: Can we have a `throw new AssertionError(e)` here?

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
```diff
@@ -0,0 +1,425 @@
...
+             */
+            tryUnregisterFromContext();
+        }
+
+        @Override
+        public void awaitFinish() {
+            stop();
+
+            // wait until handle is finished
+            lock.lock();
```
mrigger: Does this lock immediately followed by an unlock make sense?

pointhi: yes, as I had written in the comment above, it waits until no lock is held by the object. Means either some little configuration step finished, or more important, no signal handler is executed anymore.
Because we set ```isRunning``` to false before, execution of the handler would exit without calling the referenced signal handler, which means effectivly, the handler is now deactivated and we can safely unregister it.

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/intrinsics/c/LLVMSignal.java
pointhi: Yes, it's only for completeness. Normally, cleanup of the object should be handled automatically when someones stops the SignalHandler.

pointhi:
I addressed all stated issues.

To clarify why the locking mechanism is currently a bit complex, I want to state how the concurrency helpers are used:

* When someone registers a new signal handler, the old one has to be returned. Furthermore it's required to set a new handle (the last registered handle is the active one). Once the program exits, all registered handles have to be set back to default, and be deleted from the corresponding datastructures. But it has to be prevented that an old (no longer active handle) accidentally reset the signal handler while there is a new registered one. (This is the main purpose of ```isRunning```, which indicates if the current handle is the one which is managing the signal and is responsible for cleaning it up)
* The we have ```lock```, which simply is required to lock the stack as well as indicating a possible running handler. It's main purpose is to prevent multiple handlers running on the same stack as well as the handlers are finished before the stack is tear down.
* As the last lock, we have ```globalSignalHandlerLock``` which simply saves us from running multiple signal handlers at once, which could lead to some unexpected behaviour in the current implementation.

In the current implementation, it should be possible to overwrite a new signal handler while the current one is still running, and the programm shouldn't behave in an unexpected way which could lead to memory leaks or invalid signal handler definitions (like deleting an still valid signal handler)

mrigger:
Thanks for the clarifications! I think we should merge the state in, since the code is sufficiently mature and  complete. I still wonder how we could simplify the locking. Could `lock` and `globalSignalHandlerLock` be one lock? If I understand right, you use `lock` for prevent that multiple handlers run on the same stack, and `globalSignalHandlerLock` that only a single signal handler can run at once. Isn't the latter lock more conservative and ensures the stack synchronization? Could you also look how/if we can stop the main thread, so only the signal handler is executed?

PR #483, opened 2016-09-29
Some Fixes for the Binary Parser

jkreindl:


projects/uk.ac.man.cs.llvm/src/uk/ac/man/cs/llvm/ir/model/enums/AsmDialect.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package uk.ac.man.cs.llvm.ir.model.enums;
+
+public enum AsmDialect {
+
+    AT_T,
+    INTEL;
+
+    public static AsmDialect decode(long id) {
+        return values()[(int) id];
```
mrigger: Since the enum class is small, and there are (and will be) only two constants, I think it is okay to rely on the order of the enum constanst.

PR #488, opened 2016-10-03
Move the stack nuller creation to the node factory facade

mrigger:


projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/NodeFactoryFacadeImpl.java
```diff
@@ -459,4 +466,35 @@ public RootNode createStaticInitsRootNode(LLVMNode[] staticInits) {
...
+                return new LLVMFloatNuller(slot);
+            case Double:
+                return new LLVMDoubleNuller(slot);
+            case Object:
+                /**
+                 * It would be cleaner do not distinct between the frame slot kinds but the variable
+                 * type. We cannot simply set the object to null, since phis that have null and
+                 * other Object's inside escape and are allocated. We set a null address here, since
+                 * other Sulong data types that use Object are implement inefficiently anyway. In
+                 * the long term, they should have their own stack nuller.
```
chrisseaton: A few English errors here. Could instead be:

```
* It would be cleaner to not distinguish between the frame slot kinds, and use the variable
* type instead. We cannot simply set the object to null, becuase phis that have null and
* other Objects inside escape and are allocated. We set a null address here, since
* other Sulong data types that use Object are implemented inefficiently anyway. In
* the long term, they should have their own stack nuller.
```

mrigger: Thanks! I updated the PR.

PR #491, opened 2016-10-04
Intrinsic strcmp for a native address and a managed object

chrisseaton:


mrigger:
On second thought, could you still add a test case for that? You can have a look at `LLVMInteropTest` where I added test cases for other TruffleObjects-only intrinsics.

chrisseaton:
Tests added.

mrigger:
Thanks!

PR #492, opened 2016-10-04
Identity specialisation for converting to I64

chrisseaton:


chrisseaton:
Converted all simple conversions to I64 to a single node.

PR #493, opened 2016-10-04
More specialisations for address comparison

chrisseaton:


chrisseaton:
Converted to a single equal and not-equal nodes for simple types (the float stuff seems a bit more complex than that).

PR #494, opened 2016-10-04
Direct load inside a managed object

chrisseaton:


mrigger:
I rebased the PR, so we can merge it.

PR #496, opened 2016-10-04
Round up managed malloc sizes, rather than calling it an error if they aren't round

chrisseaton:


mrigger:
I rebased the PR, so we can merge it.

PR #502, opened 2016-10-10
Cleanup travis and improve special mx function to execute all basic tests at once

pointhi:


mx.sulong/mx_sulong.py
```diff
@@ -1141,6 +1148,18 @@ def checkNoHttp(args=None):
     'main-arg' : runMainArgTestCases,
 }
 
+checkCases = {
+    'gitlog' : logCheck,
```
mrigger: Can you verify that each of the code checks actually fails when it sees code violations? Do `checkoverlap` and `checkstyle` actually fail with a Python exit or just return a non-zero value? In latter case, the code check would not abort on the gate.

pointhi: I added extra checks for those functions which return an error number

mx.sulong/mx_sulong.py
```diff
@@ -1141,6 +1154,18 @@ def checkNoHttp(args=None):
...
+    'mdl' : mdlCheck,
+    'ecj' : compileWithEcjStrict,
+    'checkstyle' : checkStyle,
+    'findbugs' : findBugs,
+    'canonicalizeprojects' : canonicalizeprojects,
+    'httpcheck' : checkNoHttp,
+    'checkoverlap' : checkoverlap,
+    'clangformatcheck' : clangformatcheck
```
mrigger: Would it be also possible to add `pylint` and `eclipseformat` in this, or another PR?

PR #444, opened 2016-09-09
Add llvm-bc tests suite

SwapnilGaikwad:
- Adding 'llvm-bc' test suite for BitCode parser, equivalent to the 'llvm' test suite for .ll files.

PR #503, opened 2016-10-11
Remove the GCC and LLVM BC test suites and replace them by launching …

mrigger:
…the LL test suites with -Dsulong.TestBinaryParser=true

The GCC and LLVM .bc test suites were introduced with #444 and #449, since the bitcode parser could not parse and execute
all of the .ll GCC and LLVM test cases. Since the bitcode parser can execute all of the test cases that the textual .ll parser
can execute, the separation of the test suites and suite configuration is no longer necessary, and removed with this change.


mrigger:
@jkreindl @pointhi @SwapnilGaikwad Could one of you please have a look at this change?

jkreindl:
The changes look good, but if we want to change this anyways maybe we should think about the selection of testsuites. LLVM should be covered by Sulong, which has similar coverage to GCC but is a bit smaller. Compile also seems interesting, just to see that the parser does not crash for invalid code. I think it would be best to run the Sulong and Compile testsuites in travis4 instead of LLVM and GCC.

mrigger:
The PR does not change the test suites that are executed, it only removes existing redundancies. Maybe we can change which test suites are executed as part of another PR?

jkreindl:
Sure, I'll get to that when I have time.

SwapnilGaikwad:
Changes look good to meet too! I agree with @jkreindl .

PR #506, opened 2016-10-12
Add support for imull instruction with 2 operands

SwapnilGaikwad:
- This PR adds support for 'imull' instruction with 32bit 2 register operands. Implementation for single operand is not included.
- Test suite 'LLVMInlineAssemblyTest' is refactored to make use of available inline assembly tests in 'com.oracle.truffle.llvm.test/tests/c/inlineassembly/' directory. Currently it executes only 2 tests.
- Remove unused test cases because their equivalent tests are incorporated in the new set of test cases.
Please suggest,
- 'LLVMInlineAssemblyTest' test suite is update to make maximum use of existing test infrastructure. It is using LLVMSuiteTestCaseGenerator to create test cases. Do you have any suggestions to make it elegant?
- Test suite could be made coherent with other test suites like LLVM or GCC to make use of suite configuration etc. but may not be required in this case.


projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/LLVMPaths.java
```diff
@@ -75,6 +75,9 @@ private LLVMPaths() {
     // Interop test suites
     public static final String INTEROP_TESTS = LOCAL_TESTS + "/../interoptests";
 
+    // InlineAssembly test suite
+    public static final File INLINEASSEMBLY_TESTS = new File(LOCAL_TESTS + "/c/inlineassembly");
```
mrigger: If you have a separate test suite for inline assembly tests, then it should not be in the `LOCAL_TESTS`, since otherwise the test cases are executed by the original and the new test suite.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/inlineassembly/LLVMInlineAssemblyTest.java
```diff
@@ -29,54 +29,73 @@
...
+        for (File file : LLVMPaths.INLINEASSEMBLY_TESTS.listFiles()) {
+            testCaseFileNames.add(file.getName());
+        }
+        List<SpecificationEntry> testCaseFileSpecList = SpecificationFileReader.getFiles(testCaseFileNames, LLVMPaths.INLINEASSEMBLY_TESTS);
+        return collectIncludedFiles(testCaseFileSpecList, new LLVMSuiteTestCaseGenerator(LLVMBaseOptionFacade.testBinaryParser()));
     }
 
-    private static int run(String fileName) {
+    @Test
+    public void test() {
```
mrigger: Could you implement the test suite similar to `NWCCTestSuite` and get the reference result by compiling the test cases with GCC or Clang and then compare the executable's results with the result that Sulong produces? Then you do not have to create the `.reference_output` files by hand.

SwapnilGaikwad: Sure! Thanks!

Issue #507, opened 2016-10-12
Generate the inline assembler parser in mx


mrigger:
Currently, every change in the ATG of the inline assembler results in changes in the auto-generated parser (~500 LOC changes). The auto-generated files should not be included in the repository, similarly to the auto-generated Truffle classes for nodes and interops, since they "pollute" the project history and since they make code reviews more difficult.

It would be good if we could download Coco in the mx script, and use it to generate the parser from the ATG. Then, we could remove the checked-in auto-generated files and let each user (automatically)  generate the parser from mx.

mrigger:
Solved by #510.

PR #510, opened 2016-10-13
Remove auto-generated parser and scanner files

SwapnilGaikwad:
This PR is to address issue #507 . It includes,
- Workaround is put which wraps existing build command along with parser generation commands if necessary i.e. if .atg file is newer than generated files or any of the generated file is missing
- Download Coco.jar to the project libs if not present
- Remove tracking of auto-generated files Parser.java and Scanner.java. Also ignored the previously auto-generated files backed up by parser with .old extension
- Convert auto-generated file from Windows to Unix file format using 'dos2unix' to avoid error while 'mx checkstyle' 

Please suggest, 
- If there is an elegant way of incorporating parser files generation as part of project build command? Using some hook in mx to execute external command before building?
- Would you like to have a separate mx target like 'mx gen-parser' to generate inline assembly parser files explicitly. Current change do not check if the .frame files got updated, so one can generate files explicitly if required.


mx.sulong/mx_sulong.py
```diff
@@ -1111,6 +1115,26 @@ def checkNoHttp(args=None):
...
 
+def genInlineAssemblyParser(args=None, out=None):
+    """generate inline assembly parser and scanner if corresponding grammer is new"""
+    generatedParserDir = _inlineAssemblySrcDir + _inlineAssemblyPackageName.replace('.', '/')
+    generatedParserFile = generatedParserDir + "/Parser.java"
+    generatedScannerFile = generatedParserDir + "/Scanner.java"
+    if not isfile(generatedParserFile) or not isfile(generatedScannerFile) or os.path.getmtime(_inlineAssemblyGrammer) >= os.path.getmtime(generatedParserFile):
+        localCocoJarFile = _suite.dir + "/lib/Coco.jar"
+        if not isfile(localCocoJarFile):
+            jarFileUrls = ["http://www.ssw.uni-linz.ac.at/Coco/Java/Coco.jar"]
```
mrigger: I uploaded the file to https://lafo.ssw.uni-linz.ac.at/pub/sulong-deps/Coco.jar, so we can download it from a `https` connection.

mx.sulong/mx_sulong.py
```diff
@@ -1111,6 +1115,26 @@ def checkNoHttp(args=None):
...
+            mx.download(localCocoJarFile, jarFileUrls)
+        command = [mx.get_jdk(tag='jvmci').java, "-jar", localCocoJarFile, "-package", _inlineAssemblyPackageName, "-o", generatedParserDir, _inlineAssemblyGrammer]
+        mx.run(command)
+        #Files get generated in Windows file format. Convert them to avoid style check failure during regression testing
+        mx.run(['dos2unix', generatedParserFile, generatedScannerFile])
```
mrigger: It is unfortunate that Coco-generates files need to be reformatted. Thanks for the workaround! Could someone with a Mac verify the solution? /cc @chrisseaton @grimmerm

mx.sulong/mx_sulong.py
```diff
@@ -1111,6 +1115,26 @@ def checkNoHttp(args=None):
...
+
+def sulongBuild(args=None, out=None):
+    """custom build command to wrap inline assembly parser generation"""
+    genInlineAssemblyParser()
+    originalBuildCommand(['-p', '--warning-as-error', '--no-native', '--force-javac'])
```
mrigger: Why do we need the parameters? I do not think that we want to force users to use javac. Can we just pass `args` to `originalBuildCommand`?

mrigger:
Regarding your questions:
- I am not sure if there is a better way to register a build command.
- I think it is okay to only update the generated parser as part of `mx build`.

SwapnilGaikwad:
Thanks for your comments @mrigger , I will incorporate all of them.

mrigger:
LGTM!

Issue #512, opened 2016-10-17
Remove dos2unix dependency


mrigger:
Sulong currently has a dependency on `dos2unix` that was introduced with #510. It is part of a work around to reformat the Coco-generated inline assembler parser, in order to avoid a `mx checkstyle error` (see https://github.com/graalvm/sulong/blob/7f8cb2ab24ff1817ec3e13e1f848d398061afa94/mx.sulong/mx_sulong.py#L1131).

Adding a dependency to `dos2unix` increases the complexity of the system, and the number of external dependencies that the user has to install. It is possible to implement the reformatting with reasonable effort in Python (see http://superuser.com/questions/52044/convert-crlfs-to-line-feeds-on-linux/667743#667743). Hence, it would be better to directly implement a `dos2unix` function in mx.

PR #513, opened 2016-10-17
Remove dependency on dos2unix utility

SwapnilGaikwad:
This PR is providing fix for issue #512 . 
- The dos2unix utility is used to convert auto-generated parser files Parser.java and Scanner.java (in com.oracle.truffle.llvm.asm.amd64 project) to UNIX file format to avoid style check errors. This small code code change shall remove use of dos2unix command.
- dos2unix requirement is also removed from the Travis .yml configuration file

### DIFF
--- a/mx.sulong/mx_sulong.py
+++ b/mx.sulong/mx_sulong.py
@@ -1115,6 +1115,21 @@ def checkNoHttp(args=None):
                 exit(-1)
             line_number += 1
 
+def dos2unix(args=None):
+    """convert file from DOS to UNIX file format. Functionality equivalent to UNIX command 'dos2unix'"""
+    for fileName in args:
+        if not os.path.isfile(fileName):
+            exit("File '" + fileName + "' does not exists for dos2unix conversion.")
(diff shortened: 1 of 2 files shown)


mx.sulong/mx_sulong.py
```diff
@@ -1115,6 +1115,22 @@ def checkNoHttp(args=None):
                 exit(-1)
             line_number += 1
 
+def dos2unix(args=None):
+    """convert file from DOS to UNIX file format. Functionality equivalent to UNIX command 'dos2unix'"""
+    for fileName in args:
+        if not os.path.isfile(fileName):
```
mrigger: Could we throw an hard error here (e.g., by using `exit`)? If we pass an invalid path to the function, then we have an error in the logic of our mx script and probably always want to quit.

SwapnilGaikwad: Sure! Done!

mrigger: Thanks! LGTM!

PR #514, opened 2016-10-17
Fix a Bug with Lifetime-Analysis in the Textparser

jkreindl:


jkreindl:
@mrigger also updated the Lifetime Reference Output on lafo.

mrigger:
I executed the lifetime analysis tests locally, and the following two test cases failed:
```
projects/com.oracle.truffle.llvm.test/suites/gcc/gcc-5.2.0/gcc/testsuite/gcc.dg/ipa/ipa-icf-21.c
projects/com.oracle.truffle.llvm.test/suites/gcc/gcc-5.2.0/gcc/testsuite/gcc.dg/torture/pr52419.c

@foo expected:<begin dead [%0 ()], end dead [%0 (%3, %x, %1, %15, %4, %14, %v, %2, %7, %11, %9, %13, %5, %6, %u.i)]> but was:<begin dead [%0 ()], end dead [%0 (%x, %1, %3, %4, %5, %6, %7, %8, %9, %10, %11, %u.i, %13, %12, %14, %2, %15, %v)]>
@foo expected:<begin dead [%0 ()], end dead [%0 (%1, %p, %2, %5, %3)]> but was:<begin dead [%0 ()], end dead [%0 (%p, %1, %2, %4, %3, %5)]>
```
On the gate, a different test case (and only one) fails.

jkreindl:
It only works for me locally then. @mrigger could you restore the previous reference results on lafo while I investigate this? Otherwise no other PRs will pass Travis.

mrigger:
I generated the reference outputs based on your PR, and just pushed them to lafo. I will restart the Travis job to see whether that also works for the gate. Otherwise I will restore the old reference results.

PR #515, opened 2016-10-18
Exclude a Randomly Failing Test from Lifetime Analysis

jkreindl:
This Test generates different Lifetime-Analysis Results on some platforms, so we should exclude it. For now, the path is just hardcoded, but in the long term we should have explicit config files for this.


mrigger:
LGTM!

PR #517, opened 2016-10-20
Add support for idivl and small enhancements to inline assembly parser

SwapnilGaikwad:
This PR includes,
- Updates to mx_sulong.py file to make sure that if any of the configuration file (.atg and all .frame files) is updated then parser and scanner should be re-generated. Previously it was not checking for .frame files.
- Support idivl with two 32 bit register operands
- Support for single 32 bit register operand imull and idivl instruction
- Parser (or Parser.frame) is updated to extract number of parsing errors 
- Exit if any error occurred while parsing. e.g. idivl does not accept immediate. If we do not exit then execution fails with non-obvious error e.g. 
```
int main() {
  int arg1 = 40;
  int mul= 0;
  __asm__("imull $2;" : "=a"(mul): "a"(arg1));
  return mul;
}
```
In above case parsing error is shown on command line that ',' is expected but execution goes further resulting in ` IllegalStateException` because parser returned right operand string as '2'.

- Please suggest, 
- Currently just the error causing line is shown not the root cause as shown by clang/gcc e.g. `error: too few operands for instruction` for above example. Would you suggest to do that?
- Separate function to handle division operation due to special case of take EDX:EAX combination as dividend, any better/elegant way to do it?

### DIFF
--- a/mx.sulong/mx_sulong.py
+++ b/mx.sulong/mx_sulong.py
@@ -1132,10 +1132,14 @@ def dos2unix(args=None):
 
 def genInlineAssemblyParser(args=None, out=None):
     """generate inline assembly parser and scanner if corresponding grammer is new"""
-    generatedParserDir = _inlineAssemblySrcDir + _inlineAssemblyPackageName.replace('.', '/')
-    generatedParserFile = generatedParserDir + "/Parser.java"
-    generatedScannerFile = generatedParserDir + "/Scanner.java"
-    if not isfile(generatedParserFile) or not isfile(generatedScannerFile) or os.path.getmtime(_inlineAssemblyGrammer) >= os.path.getmtime(generatedParserFile):
+    generatedParserDir = _inlineAssemblySrcDir + _inlineAssemblyPackageName.replace('.', '/') + "/"
+    generatedFiles = [generatedParserDir + "Parser.java", generatedParserDir + "Scanner.java"]
+    configFiles = [_inlineAssemblySrcDir + "InlineAssembly.atg", _inlineAssemblySrcDir + "Parser.frame", _inlineAssemblySrcDir + "Scanner.frame", _inlineAssemblySrcDir + "Copyright.frame"]
+    isAllGeneratedFilesExists = all([isfile(fileName) for fileName in generatedFiles])
+    latestGeneratedFile = sorted(generatedFiles, key=os.path.getmtime, reverse=True)[0] if isAllGeneratedFilesExists else ""
+    latestConfigFile = sorted(configFiles, key=os.path.getmtime, reverse=True)[0]
+    #If any auto-generated file is missing or any config file is updated after last auto-generation then regenerate the files
+    if not isAllGeneratedFilesExists or os.path.getmtime(latestConfigFile) >= os.path.getmtime(latestGeneratedFile):
         localCocoJarFile = _suite.dir + "/lib/Coco.jar"
         if not isfile(localCocoJarFile):
             jarFileUrls = ["https://lafo.ssw.uni-linz.ac.at/pub/sulong-deps/Coco.jar"]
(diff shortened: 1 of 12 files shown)


mx.sulong/mx_sulong.py
```diff
@@ -1132,18 +1132,22 @@ def dos2unix(args=None):
...
-    generatedScannerFile = generatedParserDir + "/Scanner.java"
-    if not isfile(generatedParserFile) or not isfile(generatedScannerFile) or os.path.getmtime(_inlineAssemblyGrammer) >= os.path.getmtime(generatedParserFile):
+    generatedParserDir = _inlineAssemblySrcDir + _inlineAssemblyPackageName.replace('.', '/') + "/"
+    generatedFiles = [generatedParserDir + "Parser.java", generatedParserDir + "Scanner.java"]
+    configFiles = [_inlineAssemblySrcDir + "InlineAssembly.atg", _inlineAssemblySrcDir + "Parser.frame", _inlineAssemblySrcDir + "Scanner.frame", _inlineAssemblySrcDir + "Copyright.frame"]
+    isAllGeneratedFilesExists = all([isfile(fileName) for fileName in generatedFiles])
+    latestGeneratedFile = sorted(generatedFiles, key=os.path.getmtime, reverse=True)[0] if isAllGeneratedFilesExists else ""
+    latestConfigFile = sorted(configFiles, key=os.path.getmtime, reverse=True)[0]
+    #If any auto-generated file is missing or any config file is updated after last auto-generation then regenerate the files
+    if not isAllGeneratedFilesExists or os.path.getmtime(latestConfigFile) >= os.path.getmtime(latestGeneratedFile):
```
mrigger: It is great to have the parser generated only if you detect changes in the ATG or frames!

projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/asm/LLVMAMD64IdivlNode.java
```diff
@@ -0,0 +1,47 @@
...
+import com.oracle.truffle.api.dsl.NodeChildren;
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.llvm.nodes.impl.base.integers.LLVMI32Node;
+
+@NodeChildren({@NodeChild("left"), @NodeChild("rightAX"), @NodeChild("rightDX")})
+public abstract class LLVMAMD64IdivlNode extends LLVMI32Node {
+
+    @Specialization
+    public int executeI32(int left, int rightAX, int rightDX) {
+        final int registerBits = 32;
```
mrigger: Nitpicking: We could have this as a `static final` constant. If you want, you can address this with the next PR.

projects/com.oracle.truffle.llvm.asm.amd64/src/com/oracle/truffle/llvm/asm/amd64/AsmFactory.java
```diff
@@ -156,6 +158,26 @@ public void createUnaryOperation(String operation, String node) {
         this.statements.add(LLVMWriteI32NodeGen.create(opNode, slot));
     }
 
+    public void createDivisionOperatoin(String op, String divisor) {
+        FrameSlot slot = frameDescriptor.findFrameSlot(divisor);
+        LLVMI32Node divisorNode = (slot != null) ? LLVMI32ReadNodeGen.create(slot) : getImmediateNode(divisor);
+        LLVMNode statement = new LLVMI32UnsupportedInlineAssemblerNode();
```
mrigger: I think it would be easier to read if you put creating the `LLVMI32UnsupportedInlineAssemblerNode` in an else block after `op.equals("idivl")`. Otherwise, you first create a node that gets overwritten again when `op.equals("idivl")`. If you want, you can address this with the next PR.

mrigger:
I forgot: Could you please also update the instruction table in https://github.com/graalvm/sulong/blob/master/docs/INLINE-ASSEMBLER.md?

projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/NodeFactoryFacadeImpl.java
```diff
@@ -33,7 +33,6 @@
 import java.util.Map;
 import java.util.Optional;
 
-import com.oracle.truffle.llvm.parser.base.util.LLVMTypeHelper;
```
mrigger: Please try to prevent checking in files that only contain reordered imports.

PR #520, opened 2016-10-24
Refactor argon2 tests

lxp:


grimmerm:
Nice, we should reuse method pullTestSuite for all other testsuites as well.

mx.sulong/mx_sulong.py
```diff
@@ -419,6 +420,16 @@ def copytree(src, dst, symlinks=False, ignore=None):
             if not os.path.exists(d) or os.stat(s).st_mtime - os.stat(d).st_mtime > 1:
                 shutil.copy2(s, d)
 
+def pullTestSuite(library, destDir, **kwargs):
+    """downloads and unpacks a test suite"""
+    mx.ensure_dir_exists(destDir)
+    localPath = mx.library(library).get_path(True)
+    tar(localPath, destDir, **kwargs)
+    os.remove(localPath)
+    sha1Path = localPath + '.sha1'
```
mrigger: Why does this remove an `.sha1` file?

lxp: It is automatically created by mx from the metadata provided in the suite configuration. If we remove the .tar.gz file like before, we should also remove the corresponding .sha1 file.

PR #523, opened 2016-10-25
Add support for idivl instruction to return struct inline assembly 

SwapnilGaikwad:
When multiple values/registers are returned from inline assembly then the inline assembly (represented as a function call) returns a struct combining necessary values. These values are extracted by the 'extractvalue' instructions in LLVM IR followed by the call to inline assembly.
Apart from support for struct return type this commit includes,
- Updates to parser, making type as `I32` explicitly for 32 bit registers. Previously it was using same value from return type which is now struct.
- Refactoring processing of inline assembly by bitcode parser. Now using same function as that of textual parser (`createInlineAssemblerExpression` function) to process inline assembly.

Please suggest,
- idivl is calculating 32 bit quotient and 32 bit remainder separately. Is it fine to calculate them by a single function (division) and writing 8 bytes from the start address of returned structure or relying on widths would be prone to error?

### DIFF
--- a/projects/com.oracle.truffle.llvm.asm.amd64/src/com/oracle/truffle/llvm/asm/amd64/AsmFactory.java
+++ b/projects/com.oracle.truffle.llvm.asm.amd64/src/com/oracle/truffle/llvm/asm/amd64/AsmFactory.java
@@ -60,9 +60,10 @@ import com.oracle.truffle.llvm.nodes.impl.func.LLVMArgNodeFactory.LLVMI32ArgNode
 import com.oracle.truffle.llvm.nodes.impl.func.LLVMCallNode;
 import com.oracle.truffle.llvm.nodes.impl.func.LLVMInlineAssemblyRootNode;
 import com.oracle.truffle.llvm.nodes.impl.literals.LLVMSimpleLiteralNode;
-import com.oracle.truffle.llvm.nodes.impl.memory.LLVMAddressGetElementPtrNodeFactory;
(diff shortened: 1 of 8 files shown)


projects/com.oracle.truffle.llvm.asm.amd64/src/com/oracle/truffle/llvm/asm/amd64/AsmFactory.java
```diff
@@ -170,14 +183,44 @@ public void createDivisionOperatoin(String op, String divisor) {
...
             LLVMI32Node edxNode = LLVMI32ReadNodeGen.create(edxSlot);
             LLVMI32Node divisionNode = LLVMAMD64IdivlNodeGen.create(divisorNode, eaxNode, edxNode);
-            statement = LLVMWriteI32NodeGen.create(divisionNode, eaxSlot);
-            returnNode = LLVMI32ReadNodeGen.create(eaxSlot);
+            FrameSlot quotSlotEax = frameDescriptor.addFrameSlot("quotient");
+            quotSlotEax.setKind(FrameSlotKind.Int);
+            statements.add(LLVMWriteI32NodeGen.create(divisionNode, quotSlotEax));
+
+            if (retType == LLVMBaseType.STRUCT) {
+                LLVMI32Node remainderNode = LLVMAMD64ModNodeGen.create(divisorNode, eaxNode, edxNode);
```
mrigger: I think that you could compact this code by reading some information from the bitcode file, or by avoiding to create some of the frame slots. However, I think it is okay to do a refactoring the next time you implement a similar instruction, to help you identify common patterns.

mrigger:
LGTM!

PR #524, opened 2016-10-25
Support the Binary Parser in the Lifetime Testsuite

jkreindl:
This PR makes the Datastructures produced by both parsers comparable.

The reference output for the single test we excluded as part of this PR includes two basic blocks named %18 and %"18" as part of the same function block. This is an illegal situation since both names should be equivalent. This hints to either an error in the Textparser or a bug in Dragonegg.

### DIFF
--- a/projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/parser/bc/Parser.java
+++ b/projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/parser/bc/Parser.java
@@ -111,7 +111,7 @@ public class Parser {
                     UNABBREV_RECORD
     };
 
-    private static final String CHAR6 = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXJZ0123456789._";
+    private static final String CHAR6 = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789._";
 
     protected final Bitstream stream;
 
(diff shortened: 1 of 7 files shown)


projects/com.oracle.truffle.llvm.parser.bc.impl/src/com/oracle/truffle/llvm/parser/bc/impl/parser/bc/Parser.java
```diff
@@ -111,7 +111,7 @@
                     UNABBREV_RECORD
     };
 
-    private static final String CHAR6 = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXJZ0123456789._";
+    private static final String CHAR6 = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789._";
```
mrigger: I guess we should remove this table sooner or later.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/TestLifetimeAnalysisGCC.java
```diff
@@ -68,14 +80,14 @@
...
 import com.oracle.truffle.llvm.runtime.options.LLVMBaseOptionFacade;
-import com.oracle.truffle.llvm.test.FunctionVisitorIterator.LLVMFunctionVisitor;
 
 @RunWith(Parameterized.class)
 public class TestLifetimeAnalysisGCC extends TestSuiteBase {
 
     // TODO in the long term we should replace this
     private static final String[] excludedFiles = new String[]{
-                    "projects/com.oracle.truffle.llvm.test/suites/gcc/gcc-5.2.0/gcc/testsuite/gcc.dg/pr43419.c"
+                    "projects/com.oracle.truffle.llvm.test/suites/gcc/gcc-5.2.0/gcc/testsuite/gcc.dg/pr43419.c",
```
mrigger: Could you please shortly comment why you excluded the test cases as part of the PR description?

jkreindl: Done.

PR #527, opened 2016-10-28
Rewrite more parts of LLVMBitcodeVisitor to use factoryFacade - V2

jkreindl:


projects/com.oracle.truffle.llvm.parser.base/src/com/oracle/truffle/llvm/parser/base/facade/NodeFactoryFacade.java
```diff
@@ -84,7 +84,7 @@
     @Deprecated
     LLVMParserRuntime getRuntime();
 
-    LLVMExpressionNode createInsertElement(LLVMBaseType resultType, LLVMExpressionNode vector, com.intel.llvm.ireditor.lLVM_IR.Type vectorType, LLVMExpressionNode element, LLVMExpressionNode index);
+    LLVMExpressionNode createInsertElement(LLVMBaseType resultType, LLVMExpressionNode vector, LLVMExpressionNode target, LLVMExpressionNode element, LLVMExpressionNode index);
```
mrigger: I think we should not add an additional `target` node here. The `insertelement` instruction has three arguments, and that Native Sulong uses an intermediate allocation is only an implementation detail.

PR #528, opened 2016-10-31
Extend usage of LLVMParserRuntime

pointhi:
Based on the PR of @jkreindl #527 (I will rebase this PR when the other one is merged).

I extend the usage of ```LLVMParserRuntime``` into ```LLVMBitcodeInstructionVisitor```, ... which allows us to remove those repetitive ```method.getTargetDataLayout()``` which are required when calculating size, alignment, padding,...

This is also necessary work, because there are currently multiple (similar) functions which calculate alignment, size,... . With this PR I started unifying those code parts to use the calculation methods declared inside the different Types. Which means we can remove the duplicated ones currently declared inside Type Helper,...


mrigger:
Can you please rebase the change? We merged #527.

PR #530, opened 2016-11-04
Add new test infrastructure for SulongSuite

grimmerm:
With this PR we allow building Sulong Testsuite using mx.


mx.sulong/tools.py
```diff
@@ -0,0 +1,176 @@
...
+        try:
+            f = open('/dev/null', 'w')
+            return mx.run([mx_sulong.findLLVMProgram('clang'), '-S', '-emit-llvm', '-o', outputFile] + flags + [inputFile], out=f, err=f)
+        except SystemExit:
+            print 'Cannot compile %s' % (inputFile)
+        return -1
+
+    def compileReferenceFile(self, inputFile, outputFile, flags):
+        try:
+            f = open('/dev/null', 'w')
```
mrigger: That only works for Unix system. Could you solve that in a platform-independent way?

mx.sulong/tools.py
```diff
@@ -0,0 +1,176 @@
...
+        return -1
+
+class Opt(Tool):
+    def __init__(self, name, passes):
+        self.name = name
+        self.supportedLanguages = [ProgrammingLanguage.LLVMIR]
+        self.passes = passes
+
+    def supports(self, language):
+        print 'supports? %s' %(language.name)
```
mrigger: Is that left-over debug output or intended?

mx.sulong/tools.py
```diff
@@ -0,0 +1,176 @@
...
+
+Runtime.LLI = LLIRuntime()
+
+
+def getFileExtension(f):
+    _, ext = os.path.splitext(f)
+    return ext[1:]
+
+def getOutputName(inputFile, outputDir, tool, optimization, target):
+    base, _ = os.path.splitext(inputFile)
```
mrigger: It would be nice if you could add a one-line `"""` comment to explain what these functions do, similar to how we document in the other mx files

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/alpha/SulongSuite.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import com.oracle.truffle.llvm.LLVM;
+import com.oracle.truffle.llvm.runtime.options.LLVMBaseOptionFacade;
+import com.oracle.truffle.llvm.tools.util.ProcessUtil;
+
+@RunWith(Parameterized.class)
+public class SulongSuite {
```
mrigger: Can we add this test suite to our gate? Then we have the new code already tested.

mx.sulong/tools.py
```diff
@@ -0,0 +1,176 @@
...
+
+    outputDir = os.path.dirname(outputPath)
+    if not os.path.exists(outputDir):
+        os.makedirs(outputDir)
+    return '%s.%s' % (outputPath, target.exts[0])
+
+def multicompileFile(inputFile, outputDir, tools, flags, optimizations, target, optimizerTools):
+    base, ext = os.path.splitext(inputFile)
+    lang = ProgrammingLanguage.lookup(getFileExtension(inputFile))
+    for tool in tools:
```
mrigger: I think using generator expressions here would make the code more readable.

mx.sulong/tools.py
grimmerm: yeah, will use f = open(os.devnull, 'w')

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/alpha/SulongSuite.java
grimmerm: I would like to wait until we remove the old SulongSuite by the new one.

PR #534, opened 2016-11-07
Remove trivial LLVM IR tests from the Sulong test suite

mrigger:
LLVM IR test cases are difficult to maintain, since they are platform dependent.
Thus, this change removes some very simple test cases that no longer add coverage.


grimmerm:
Great! Thx for the change!

PR #539, opened 2016-11-09
Use new compile infrastructure for LLVM test suite

lxp:


mx.sulong/mx_sulong.py
```diff
@@ -1221,4 +1236,5 @@ def sulongBuild(args=None):
     'su-get-llvm-program' : [getLLVMProgramPath, ''],
     'su-get-gcc-program' : [getGCCProgramPath, ''],
     'su-compile-sulong-suite' : [compileSulongSuite, ''],
+    'su-compile-llvm-suite' : [compileLLVMSuite, ''],
```
mrigger: Would it make sense if we had a `su-compile` command that takes the arguments `sulong` and `llvm`? Otherwise we just keep expanding the mx command with each newly supported test suite.

mx.sulong/suite.py
```diff
@@ -215,6 +215,14 @@
       ],
       "sha1" : "5552052e53fcd7fe40c558866c9cd51027c17322",
     },
+    "LLVMTESTSUITE" : {
```
mrigger: Can we change that to `LLVM_TEST_SUITE` to match the naming convention in this file?

mx.sulong/suite.py
```diff
@@ -215,6 +215,14 @@
...
+      "path" : "tests/test-suite-3.2.src.tar.gz",
+      "urls" : [
+        "https://lafo.ssw.uni-linz.ac.at/pub/sulong-deps/test-suite-3.2.src.tar.gz",
+        "http://llvm.org/releases/3.2/test-suite-3.2.src.tar.gz",
```
mrigger: We cannot have any `http` links due to security reasons. Can you remove the link from `llvm.org`? There is a `http-check` command but I think it does not check the `suite.py` file. Can you add the check for this file?

lxp: The `http` link is okay in this situation, as mx verifies the sha1 hash of the download. I will still remove the `http` link as we don't really need it.

mx.sulong/suite.py
lxp: I will change that.

PR #541, opened 2016-11-09
Metadata cleanup

pointhi:
* removing not used visitors (use interface with default implementations)
* add visitor pattern to Metadata Nodes
* restructure big parts of ```LLVMMetadata``` to be easier to extend, easier to read,...
* add lots of comments

I found out my implementation had some bigger flaws. Especially with bitfields which are made using add, sub, shift,... operators. But the most simple case, variable inside struct should work in most cases.


projects/com.oracle.truffle.llvm.parser.base/src/com/oracle/truffle/llvm/parser/base/model/metadata/subtypes/MSTName.java
```diff
@@ -0,0 +1,9 @@
+package com.oracle.truffle.llvm.parser.base.model.metadata.subtypes;
+
+import com.oracle.truffle.llvm.parser.base.model.blocks.MetadataBlock.MetadataReference;
+
+public interface MSTName {
```
mrigger: What's MST? Could we use the full form of the abbreviation (if it is one?)

pointhi: I used it instead of ```MetadataSubtype``` (couldn't find a better name)

PR #540, opened 2016-11-09
Add stdout redirection feature for Sulong testing

grimmerm:
Sulong's test harness for its testsuites compares the stdout of a reference execution with Sulong's execution.
This commit adds a feature that allows capturing the stdout of the Sulong execution (Java and native prints) and the whitelist feature.

PR #542, opened 2016-11-10
Add compileWithEcjStrict and finbugs to Sulong gate

grimmerm:
These checks were removed during hunting a compilation issue but should be in the gate.


mrigger:
Note: Introduced in #540.

PR #548, opened 2016-11-15
Add some regression tests

pointhi:
This are test cases which check for regressions in the metadata parser.

Some additional information can be found here: https://github.com/pointhi/sulong/wiki/Regression-Test-Codes

There is the problem that metadata parsing only works with enabled debug symbols, as well as some test cases need activated optimisations. Either we specify such constrains for those specific test cases, or we add them precompiled in bitcode or llvm ir format so they are always executed as intended.


grimmerm:
These files need to be compiled with mx in the long run.
How do you currently compile the files?

mrigger:
@grimmerm: I think the test cases are currently compiled **without** the `-g` flag, and with the compilation logic of the `SulongSuite` since the test cases are added to the `test` subdirectory.

grimmerm:
ok, that also means that currently nobody tests the metadata parser, right?

In the long run, we might want to always add -g

grimmerm:
Closed because there was no progress for more than one month.
/cc @pointhi

PR #549, opened 2016-11-15
Force livetimeanalysis to always compile to bitcode

pointhi:


grimmerm:
@pointhi, can you, as a next step, use compiled files instead of compiling them on demand?

PR #552, opened 2016-11-16
Move file Parser/Writer of Lifetime Analysis into own class

pointhi:
Move the reference file handling logic of ```TestLifetimeAnalysisGCC``` into a new class, and use a event driven approach to connect it to the parser.


projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/util/LifetimeFileParserEvents.java
```diff
@@ -0,0 +1,17 @@
+package com.oracle.truffle.llvm.test.util;
+
+public interface LifetimeFileParserEvents {
```
mrigger: Could we rename that to `LifetimeFileParserListener`?

projects/com.oracle.truffle.llvm.parser.impl/src/com/oracle/truffle/llvm/parser/impl/lifetime/LLVMLifeTimeAnalysisVisitor.java
```diff
@@ -1,392 +0,0 @@
...
-import com.oracle.truffle.llvm.parser.impl.LLVMPhiVisitor.Phi;
-import com.oracle.truffle.llvm.parser.impl.LLVMReadVisitor;
-import com.oracle.truffle.llvm.runtime.options.LLVMOptions;
-
-/**
- * This class determines which variables are dead after each basic block. It applies an iterative
- * data-flow analysis to determine the lifetimes.
- *
- */
-public final class LLVMLifeTimeAnalysisVisitor {
```
mrigger: We cannot remove this class yet, since it is essential for peak performance (when using the textual parser). Only when we remove the textual parser completely, we can also remove it. This is also why the gate fails with a Graal bailout.

PR #553, opened 2016-11-16
Run Sulong suite using new test infrastructure

grimmerm:
Replace old SulongTestSuite and use new infrastructure.


mx.sulong/mx_sulong.py
```diff
@@ -573,10 +573,9 @@ def runLLVMTestCases(args=None):
 
 def runTruffleTestCases(args=None):
     """runs the Sulong test suite"""
-    ensureLLVMBinariesExist()
-    ensureDragonEggExists()
+    testsuites.compileSuite(['sulong'])
     vmArgs, _ = truffle_extract_VM_args(args)
-    return unittest(getCommonUnitTestOptions() + ['-Dsulong.ExecutionCount=1000'] + vmArgs + ["com.oracle.truffle.llvm.test.SulongTestSuite"])
+    return unittest(vmArgs + ["com.oracle.truffle.llvm.test.alpha.SulongSuite"])
```
grimmerm: I intentionally do not pass any options here.
I feel like we should only pass the options that we actually need to run the suite.
/cc @mrigger

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/alpha/BaseSuite.java
```diff
@@ -117,7 +118,15 @@ public void test() throws Exception {
             System.err.println(testDiscoveryPath);
             whiteListFilter = p -> !whiteList.contains(p) && p.startsWith(new File(suiteDir.toString(), testDiscoveryPath).toPath());
         }
-        return collectTestCases(suiteDir, whiteListFilter);
+        Collection<Object[]> testCases = collectTestCases(suiteDir, whiteListFilter);
+        if (testCases.size() != whiteList.size()) {
+            // we are missing test cases that are on the whitelist!
```
grimmerm: This test harness only runs tests on the whitelist.
Also, it ensures that all tests on the whitelist pass.

mx.sulong/mx_sulong.py
mrigger: If we remove the execution count, we do not longer test that Graal can still compile our nodes. This is a problem, as long as we do not integrate such a downstream test.

mx.sulong/mx_sulong.py
```diff
@@ -573,10 +573,9 @@ def runLLVMTestCases(args=None):
 
 def runTruffleTestCases(args=None):
     """runs the Sulong test suite"""
-    ensureLLVMBinariesExist()
-    ensureDragonEggExists()
```
lxp: We still need to call `ensureDragonEggExists()` somewhere, but maybe we should move it to `tools.py` `GCCCompiler.run`.

projects/com.oracle.truffle.llvm.test/src/com/oracle/truffle/llvm/test/alpha/BaseSuite.java
```diff
@@ -80,6 +80,7 @@ public void test() throws Exception {
             CaptureOutput.startCapturing();
             int sulongResult = -1;
             try {
+                assert candidate.toAbsolutePath().toFile().exists() : "File " + candidate.toAbsolutePath().toFile() + " does not exist.";
```
mrigger: Maybe we could make an explicit check out of that, and throw an exception if the file does not exists??

mx.sulong/mx_sulong.py
grimmerm: OK, I agree.
But rather than running tests 1000x I suggest:
graal.TruffleCompileImmediately=true in combination with TruffleBackgroundCompilation=false

This actually tests more because you cannot run a method without compilation.
I need to find out how much time this takes.

mx.sulong/testsuites.py
```diff
@@ -32,7 +32,7 @@ def compileLLVMSuite():
     print("Compiling LLVM Suite reference executables ", end='')
     tools.printProgress(tools.multicompileRefFolder(_llvmSuiteDir, _cacheDir, [tools.Tool.CLANG], ['-Iinclude'], excludes=excludes))
     print("Compiling LLVM Suite with -O0 ", end='')
-    tools.printProgress(tools.multicompileFolder(_llvmSuiteDir, _cacheDir, [tools.Tool.CLANG], ['-Iinclude'], [tools.Optimization.NONE], tools.ProgrammingLanguage.LLVMIR, excludes=excludes))
+    tools.printProgress(tools.multicompileFolder(_llvmSuiteDir, _cacheDir, [tools.Tool.CLANG], ['-Iinclude'], [tools.Optimization.NONE], tools.ProgrammingLanguage.LLVMBC, excludes=excludes))
```
mrigger: It was introduced by a previous commit, but some of the constants use the notation style `I_AM_A_CONSTANT` and others `IAMACONSTANT`. Maybe we should change these (including `LLVMIR` and `LLVMBC`) to use the `I_AM_A_CONSTANT` notation?

grimmerm:
Closed as duplicate:
https://github.com/graalvm/sulong/pull/569

PR #564, opened 2016-11-21
Update pip dependency

pointhi:


jkreindl:
@pointhi could you please close this PR. #565 solves the same problem.

PR #569, opened 2016-11-22
Run Sulong suite using new test infrastructure (Rebased against master)

lxp:
Replace old SulongTestSuite and use new infrastructure.

Same pull request as https://github.com/graalvm/sulong/pull/553, but rebased against master.


.travis.yml
```diff
@@ -57,5 +57,10 @@ matrix:
     jdk: oraclejdk8
   - env: TEST_COMMAND='mx su-travis-argon2' CLANG='suite version (3.2)'
     jdk: oraclejdk8
+  - env: TEST_COMMAND='mx su-travis-tests sulong' CLANG='suite version (3.2)'
+    jdk: oraclejdk8
+  allow_failures:
+  - env: TEST_COMMAND='mx su-travis-tests sulong' CLANG='suite version (3.2)'
```
mrigger: Why do we allow failures here?

mx.sulong/testsuites.py
```diff
@@ -22,11 +24,18 @@
 def compileSulongSuite():
     print("Compiling Sulong Suite reference executables ", end='')
     tools.printProgress(tools.multicompileRefFolder(_sulongSuiteDir, _cacheDir, [tools.Tool.CLANG], ['-Iinclude', '-lm']))
-    print("Compiling Sulong Suite with -O0 ", end='')
-    tools.printProgress(tools.multicompileFolder(_sulongSuiteDir, _cacheDir, [tools.Tool.CLANG, tools.Tool.GCC], ['-Iinclude', '-lm'], [tools.Optimization.O0], tools.ProgrammingLanguage.LLVMBC, optimizers=[tools.Tool.BB_VECTORIZE]))
-    print("Compiling Sulong Suite with -O1/2/3 ", end='')
-    tools.printProgress(tools.multicompileFolder(_sulongSuiteDir, _cacheDir, [tools.Tool.CLANG, tools.Tool.GCC], ['-Iinclude', '-lm'], [tools.Optimization.O1, tools.Optimization.O2, tools.Optimization.O3], tools.ProgrammingLanguage.LLVMBC))
-    # MG: compared to the old test suite we do not run the ll files
+    print("Compiling Sulong Suite with clang -O0 ", end='')
```
mrigger: It is not directly related to this PR, but I think that we should hide this debug/status information behind a `-v` flag

mx.sulong/testsuites.py
```diff
@@ -61,8 +69,25 @@ def compileSuite(args=None):
...
+
+def travisRunSuite(args=None):
+    """executes all the test suites or selected ones (see -h or --help)"""
+    vmArgs, otherArgs = mx_sulong.truffle_extract_VM_args(args)
+    parser = argparse.ArgumentParser(description="Compiles all or selected test suites.")
+    parser.add_argument('suite', nargs='*', help=' '.join(testSuites.keys()), default=testSuites.keys())
+    parsedArgs = parser.parse_args(otherArgs)
+
+    tasks = []
+    with mx_gate.Task('BuildJavaWithJavac', tasks) as t:
```
mrigger: I think that we should use the `mx gate` command for this in the future. What I did not know when I implemented the `su-travis*` command is that you can also select jobs through arguments.

mx.sulong/testsuites.py
lxp: `-v` currently gives even more details (i.e. enables compiler output)

lxp: I think we shouldn't hide that information, because it is already rather minimal. Also the progress dots are necessary to workaround the travis timeout.

Issue #375, opened 2016-08-10
Support rdtsc instruction in inline assembler (multiple return registers)


lxp:
Support the following inline assemble snipped on amd64 (as used by argon2 benchmark):

``` c
static uint64_t rdtsc(void) {
    uint64_t rax, rdx;
    __asm__ __volatile__("rdtsc" : "=a"(rax), "=d"(rdx) : :);
    return (rdx << 32) | rax;
}
```

To reproduce, run argon2 benchmark: `mx su-bench-argon2`

```
-- line 1 col 2: invalid InlineAssembly
Exception in thread "main" java.lang.AssertionError: STRUCT
    at com.oracle.truffle.llvm.parser.factories.NodeFactoryFacadeImpl.createInlineAssemblerExpression(NodeFactoryFacadeImpl.java:388)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visitFunctionCall(LLVMVisitor.java:742)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visitFunctionCall(LLVMVisitor.java:1036)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visitNamedMiddleInstruction(LLVMVisitor.java:782)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visitMiddleInstruction(LLVMVisitor.java:703)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visitInstruction(LLVMVisitor.java:691)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visitBasicBlock(LLVMVisitor.java:670)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.getFunctionBlockStatements(LLVMVisitor.java:536)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visitFunction(LLVMVisitor.java:516)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.visit(LLVMVisitor.java:297)
    at com.oracle.truffle.llvm.parser.impl.LLVMVisitor.getMain(LLVMVisitor.java:263)
    at com.oracle.truffle.llvm.LLVM.parseString(LLVM.java:254)
    at com.oracle.truffle.llvm.LLVM$1.lambda$parse$1(LLVM.java:142)
    at com.oracle.truffle.llvm.SulongLibrary.readContents(SulongLibrary.java:88)
[...]

PR #589, opened 2016-11-25
Add rdtsc instruction to inline assembly

SwapnilGaikwad:
This PR is to add support for rdtsc instruction in inline assembly. It shall also address the issue #375 .
This PR includes the following changes
- Changes to inline assembly parser generation file to incorporate detecting rdtsc instruction and little refactoring
- Implementation of rdtsc using `System.nanoTime()` routine
- Little refactoring to handle struct return types. Moved creation of nodes to allocate structure and populate with appropriate result at a single place (useful for `idivl` and `rdtsc` instructions)
- Add test to ensure correct parsing of `rdtsc` instruction

Implementation of RDTSC instruction returns current nanotime value instead of reading value stored in TSC register. I could not find any simpler way to retrieve this information in Java. Also some of the aricles encouraged using nanotime value to do elapsed time measurements [1][2]. 
As the idea of using rdtsc is to identify elapsed time this alternative can be acceptable. Therefore the test `inlineassembly025.c` only checks for correct parsing of rdtsc instruction, it does not compare the value with the c execution as they will not be the same.
Please let me know if this is alright.

[1] http://btorpey.github.io/blog/2014/02/18/clock-sources-in-linux/
[2] https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/asm/LLVMAMD64RdtscNode.java
+++ b/projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/asm/LLVMAMD64RdtscNode.java
@@ -29,11 +29,13 @@
  */
 package com.oracle.truffle.llvm.nodes.impl.asm;
 
+import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.llvm.nodes.impl.base.integers.LLVMI64Node;
 
 public abstract class LLVMAMD64RdtscNode extends LLVMI64Node {
 
+    @TruffleBoundary
     @Specialization
     public long executeRdtsc() {
         return System.nanoTime();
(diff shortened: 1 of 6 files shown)


projects/com.oracle.truffle.llvm.nodes.impl/src/com/oracle/truffle/llvm/nodes/impl/asm/LLVMAMD64RdtscNode.java
```diff
@@ -0,0 +1,41 @@
...
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.nodes.impl.asm;
+
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.llvm.nodes.impl.base.integers.LLVMI64Node;
+
+public abstract class LLVMAMD64RdtscNode extends LLVMI64Node {
+
+    @Specialization
```
mrigger: You need to add a `@TruffleBoundary` here, since otherwise the compiler would try to continue to inline through the Java library call.

PR #607, opened 2016-11-28
Remove unnecessary filter on folders containing an executable during …

grimmerm:
…test discovery


grimmerm:
I think we won't fix that.
The filter(isExecutable) is important. It ensures that we only have folders with working executables, which we need to run the discovery mode.

If a future test suite does not have executables, it is probably best to overwrite the lookup of test candidates.

PR #609, opened 2016-11-28
Fix relpath issue in Sulong's compiler tools and move sulong/test/cac…

grimmerm:
…he to sulong/cache


grimmerm:
That is tricky if it is the include file that changed. But yes, that would be awesome to have. But not in this PR.

mrigger:
Right, sorry for the confusion! We already have such a check as you said, and having a checksum on the bitcode level would defeat the purpose of caching. Just forget my suggestion. ;)

A detection on some files that have an influence on all test files could be an interesting addition in the future, as you suggested.

PR #613, opened 2016-11-29
Update import versions, and use new mx features in `llvm.pipe` project

rschatz:


mrigger:
Could you please also update the `README.md`? You can copy the changes from #587.

rschatz:
Whoops, missed that. Fixed.

PR #615, opened 2016-11-29
su-build command to build files

chrisseaton:
I'll switch to a native project for the Truffle library later, but for now I just need a way to make sure the `.bc` files are built.


grimmerm:
I do not like these fast solutions in the Sulong project. Given that we have a proper solution for that, I would prefer a clean implementation. Finding and doing these necessary refactorings will take much more time when done later on.

chrisseaton:
I can do a native project. But at the moment with the current system I've got no way to force this files to be built. This is fixing a bug in the current system so I can use it, until I can get around to improving it with a native project.

grimmerm:
see https://github.com/graalvm/sulong/pull/617

PR #616, opened 2016-11-29
DRAFT using makefiles instead of overwriting mx build

grimmerm:


grimmerm:
Closed, solution in: https://github.com/graalvm/sulong/pull/617

PR #625, opened 2016-12-01
Fix toi64node

mickjordan:


mickjordan:
My commit message all start with an uc character, so what is the problem? It's a ridiculous check IMHO anyway

grimmerm:
I did rebase your PR for you, which removed all old commits and resolves your gate failure.

PR #628, opened 2016-12-05
Output LLVM IR from parsed datastructures and allow creating those LLVM IR datastructures using java

pointhi:
This is some initial code, so we are able to generate ```LLVM IR``` using java written test definitions.

The required implementation work to make this possible only consists of two parts:
* A factory to simplify generating Instructions on the fly (Based on methods of ```InstructionGenerator```)
* A visitor which generates ```LLVM IR``` from a ```FunctionDefinition```

The remaining implementation details are already present in sulong. There was only the need to do some little altering of existing data structures, to simplify the generator code. (like changing some toString methodes to return the ```LLVM IR``` equivalent type name.


projects/com.oracle.truffle.llvm.parser.api/src/com/oracle/truffle/llvm/parser/api/model/enums/BinaryOperator.java
```diff
@@ -61,12 +61,18 @@ public static BinaryOperator decode(int opcode, boolean isFloatingPoint) {
...
+    BinaryOperator(int fpmap, String name) {
         this.fpmap = fpmap;
+        this.name = name;
     }
 
     private BinaryOperator fp() {
         return fpmap < 0 ? null : values()[fpmap];
     }
+
+    public String getName() {
```
jkreindl: I think this should rather be toString()

projects/com.oracle.truffle.llvm.writer/src/com/oracle/truffle/llvm/writer/facades/FunctionV32Writer.java
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+
+        private static String createSymbolLabel(InstructionBlock s) {
+            return "label %" + createBranchLabel(s);
+        }
+
+        @Override
+        public void visit(AllocateInstruction allocate) {
+            List<String> sb = new ArrayList<>();
```
jkreindl: Could you please use a StringBuilder or, better yet, String.format() instead of lists in these methods.

projects/com.oracle.truffle.llvm.parser.api/src/com/oracle/truffle/llvm/parser/api/model/symbols/Symbol.java
```diff
@@ -37,6 +37,10 @@
 
     boolean hasName();
 
+    default String getName() {
+        return "<anon>";
```
jkreindl: Could you make this static and replace ValueSymbol.ANON with it?

projects/com.oracle.truffle.llvm.parser.api/src/com/oracle/truffle/llvm/parser/api/model/symbols/instructions/AllocateInstruction.java
```diff
@@ -81,4 +83,9 @@ public static AllocateInstruction fromSymbols(Symbols symbols, Type type, int co
         inst.count = symbols.getSymbol(count, inst);
         return inst;
     }
+
+    @Override
+    public String toString() {
+        return String.format("%s = %s %s, align %d", getName(), LLVMIR_LABEL, getPointeeType(), 1 << (getAlign() - 1));
```
jkreindl: In 'alloca' the explicit alignment is optional.

jkreindl:
@mrigger I agree that toString cannot always generate valid IR. However, I still think that having the toString methods generate at least partially valid IR would make debugging much easier. Perhaps we could have the toString methods return only the IR definition of a symbol and then use a visitor to take care of formatting, name associtation and printing.

mrigger:
@jkreindl sounds good! It is important to have good `toString` output, but it should be for the programmer's sake, and not for parsing it. Otherwise, we will have to stick to that `toString` implementation (see Effective Java "Item 10: Always override `toString`").

projects/com.oracle.truffle.llvm.parser.bc/src/com/oracle/truffle/llvm/parser/bc/parser/listeners/module/ModuleV32.java
```diff
@@ -55,14 +56,20 @@ protected void createFunction(long[] args) {
...
+        int initialiser = (int) args[2];
+        long linkage = args[3];
+        int align = (int) args[4];
 
-        generator.createGlobal(type, isConstant, initialiser, align, linkage);
+        long visibility = Visibility.DEFAULT.getEncodedValue();
+        if (args.length >= 7) {
+            visibility = args[6];
+        }
+        // Checkstyle: stop magic number name check
```
jkreindl: Yes it should.

projects/com.oracle.truffle.llvm.parser.api/src/com/oracle/truffle/llvm/parser/api/model/enums/CompareOperator.java
```diff
@@ -76,4 +76,21 @@ public static CompareOperator decode(long opcode) {
         }
         return null;
     }
+
+    private final boolean fp;
```
grimmerm: `fp` is not a good name.

pointhi:
@grimmerm, I think you can start reviewing this PR.

At the current state, some initial tests were implemented, which shows how the ```InstructionGeneratorFacade``` works and how the LLVM IR code looks like when we using the new ```toString``` methods.

PR #630, opened 2016-12-06
Support Priorities for Constructor and Destructor Attributes

jkreindl:


projects/com.oracle.truffle.llvm.parser.bc/src/com/oracle/truffle/llvm/parser/bc/LLVMBitcodeVisitor.java
```diff
@@ -348,10 +359,74 @@ private LLVMExpressionNode getGlobalVariable(GlobalValueSymbol global) {
...
+            final StructureConstant structureConstant = (StructureConstant) arrayConstant.getElement(i);
+            final Integer priorityConstant = LLVMSymbolResolver.evaluateIntegerConstant(structureConstant.getElement(0));
+            assert priorityConstant != null;
+            sortedStructors.add(priorityConstant, functionCall);
+        }
+
+        return sortedStructors.getAll();
+    }
+
+    private static final class SortedLLVMNodeList {
```
jkreindl: Java's PriorityQueue does not support priorities separate from the data objects and the specification of [toArray()](https://docs.oracle.com/javase/8/docs/api/java/util/PriorityQueue.html#toArray--) explicitly states that the elements do not follow a particular order. Considering these drawbacks I decided that an explicit implementation would require the least code.

mrigger: Thanks for the clarification! Would `Arrays.sort` be applicable to replace the list class? I think that maintaining our own list data structures is error-prone (maybe because I hardly ever achieve to implement a bug-free list implementation :smile:). If you think doing so is more complicated, we better keep the list implementation.

jkreindl: Arrays.sort() also does not support sorting keys separate from data values, so we would not get rid of the additional element class anyways. The list just seemed a bit more expressive than a solution with Arrays.sort(). Nonetheless, if you think its preferable I can do that too. I'll replace it tomorrow.

grimmerm: @jkreindl, please do not introduce a new collection implementation. As Manuel stated, this is error-prone.
If I understand your code correctly, you want to assign a priority to an `LLVMNode` and sort by this priority.

One trivial solution could be that you have 
```
class PriorityElement {
  final LLVMNode node;
  final int priority;
}

```
and have this in a sorted Collection. Use a comparator to sort by priorities.

PR #631, opened 2016-12-07
Sulong cleanup

grimmerm:
This PR contains a bunch of cleanup commits, which you should review commit by commit.

- Make `AllocationResult` a value type. There is no reference comparison of `AllocationResults` in the code nor do they escape. `ValueType` annotation can be applied.
- `System.arraycopy` can be dangerous, i've replaced many of those by a for-loop. AFAIK, if the array length and element type information is not available, it is much slower than a simple for-loop.
- Flatten `LLVM80BitFloat` and `LLVMIVarBit` class hierarchy: please always specialize on the actual type (not an abstract super type or interface;  specializations on the actual type avoid vcalls etc.)
- Flatten vector implementation and most importantly remove lambdas and generics. AFAIK, lambdas cause a bad startup. Hence, I've removed them. Also, avoid generics. In the previous implementation we would have constant boxing/unboxing.
- Vectors can by `ValueTypes`. 

Summary of small changes:
- adding `ExplodeLoop` annotation to all loops where it is obviously missing (and causes an escaping frame)
- adding dimensions to all `CompilationFinal` array fields (javadoc states that dimension information should be provided)
- adding `TruffleBoundary` to all methods that look dangerous for PE


projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/vector/LLVMDoubleVector.java
```diff
@@ -49,48 +55,90 @@ public static LLVMDoubleVector fromDoubleArray(LLVMAddress target, double[] vals
...
+        for (int i = 0; i < values.length; i++) {
+            values[i] = getValue(i);
+        }
+        return values;
+    }
+
     public LLVMDoubleVector add(LLVMAddress addr, LLVMDoubleVector right) {
-        return performOperation(addr, right, (a, b) -> a + b);
+        LLVMAddress currentAddr = addr;
+        for (int i = 0; i < nrElements; i++) {
```
mrigger: Removing the lambdas makes sense. However, we duplicated lots of code now. Could we instead use anonymous classes or use some other abstraction?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/cast/LLVMToI8VectorNode.java
```diff
@@ -39,29 +39,28 @@
...
 
         @Specialization
         public LLVMI8Vector executeI8Vector(LLVMAddress target, LLVMI16Vector from) {
-            return executeI8VectorBody(target, from);
+            return getI8(target, from);
+        }
+
+        @ExplodeLoop
+        private static LLVMI8Vector getI8(LLVMAddress target, LLVMI16Vector from) {
+            int length = from.getLength();
```
mrigger: Would be possible to have a common method for these casts?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/x86/LLVMX86_64BitVAStart.java
```diff
@@ -231,6 +225,7 @@ private static LLVMRuntimeType getArgumentType(Object arg) {
         return type;
     }
 
+    @ExplodeLoop
```
mrigger: Why do we need the `@ExplodeLoop` here? There are no node arrays which the PE should explode, or are there?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/func/LLVMCallNode.java
```diff
@@ -314,36 +295,35 @@ public LLVMFunctionCallChainStartNode(LLVMExpressionNode functionCallNode, LLVMF
...
+                return doExecute(frame, foreignExecute, function, args, toLLVM, chain.returnType);
             }
         }
 
-    }
-
-    public static Object[] convertToPrimitiveArgs(Object[] arguments) {
-        Object[] newArguments = new Object[arguments.length - LLVMCallNode.ARG_START_INDEX];
-        System.arraycopy(arguments, LLVMCallNode.ARG_START_INDEX, newArguments, 0, newArguments.length);
-        CompilerAsserts.compilationConstant(arguments.length);
```
mrigger: I do not think that it is a good idea to replace the `System.arraycopy` calls with the more verbose for loop. The `CompilerAsserts.compilationConstant` ensures that arguments.length is final, so Graal should be able to optimize the copy away.

PR #634, opened 2016-12-14
Add mx benchmark command support

grimmerm:
Running with sulong:
`mx benchmark csuite:* -- --jvm server --jvm-config graal-core`

Running with clang:
`mx benchmark csuite:* -- --clang -- -O3`

### DIFF
--- a/mx.sulong/mx_sulong_benchmarks.py
+++ b/mx.sulong/mx_sulong_benchmarks.py
@@ -1,0 +1,176 @@
+
+import re
+import mx, mx_benchmark, mx_sulong
+import os
+from os.path import join
+from mx_benchmark import JavaBenchmarkSuite
+
+
+def _benchmarksDirectory():
+    return join(os.path.abspath(join(mx.suite('sulong').dir, os.pardir)), 'sulong-benchmarks')
+
+class SulongBenchmarkSuite(JavaBenchmarkSuite):
+    def group(self):
+        return 'Graal'
+
+    def subgroup(self):
+        return 'sulong'
+
+    def name(self):
+        return 'csuite'
+
+    def benchmarkList(self, bmSuiteArgs):
+        benchDir = _benchmarksDirectory()
+        return [f for f in os.listdir(benchDir) if f.endswith('.c') and os.path.isfile(join(benchDir, f))]
+
+    def getSulongEngine(self, bmSuiteArgs):
+        return self.parseCommandLineArgs(bmSuiteArgs)[0][0]
+
+    def getNativeArgs(self, bmSuiteArgs):
+        return self.parseCommandLineArgs(bmSuiteArgs)[1]
+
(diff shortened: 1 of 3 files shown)


mx.sulong/mx_sulong_benchmarks.py
```diff
@@ -0,0 +1,257 @@
...
+          mx_benchmark.StdOutRule(_benchHigherScoreRegex(), {
+                "vm": "sulong",
+                "benchmark": ("<benchmark>", str),
+                "metric.name": "score",
+                "metric.type": "numeric",
+                "metric.value": ("<score>", float),
+                "metric.score-function": "id",
+                "metric.better": "higher",
+                "metric.iteration": 0,
+                "host-vm": "native",
```
gilles-duboscq: So far we have the reverse convention: for VMs/executions where there is no concept of "host" vs. "guest" VM (e.g., GNU R), then it's the "guest" which is `none`.

mx.sulong/mx_sulong_benchmarks.py
```diff
@@ -0,0 +1,257 @@
...
+        elif ext == '.cpp':
+            mx.run(['g++'] + bmSuiteArgs + ['-lm', '-lgmp', '-O3'] + [inputFile], out=f, err=f)
+        else:
+            exit(ext + " is not supported!")
+
+        myStdOut = mx.OutputCapture()
+        retCode = mx.run(['./a.out'], out=myStdOut, err=f)
+        return [retCode, myStdOut.data]
+
+class CLANGBenchmarkSuite(NativeBenchmarkSuite):
```
gilles-duboscq: The idea was that there is one `BenchmarkSuite` per benchmark suite, this `BenchmarkSuite` then should delegate to other objects for execution.
To take a java-world example, one "DaCapo" `BenchmarkSuite` class that delegates to a `JavaVM` class for execution. This benchmark suite then has options (`--jvm` and `--jvm-config`) that allow it to pick the right `JavaVM`. Then the benchmark suite composes the dacapo-specific part for the command line (e.g., `["-jar", "dacapo.jar", "fop"]`) and asks the `JavaVM` to launch a java VM with those args. `JavaVM` can build the rest of the command line to get "HotSpot configured with Graal", or "HotSpot configured with C2" etc.

So you could have such as `NativeVM` class with implementations for GCC, clang and Sulong.

In the case of hosted VMs such as Sulong you need an additional layer of delegation since such a `SulongVM` would have to delegate to a `JavaVM` for the Java part.

gilles-duboscq:
This looks better with one `BenchmarkSuite`, thanks for making the changes!

I still think we should have a better way to do this for the "sulong" engine around the detection of `host-vm` and `guest-vm` and in the selection of the host VM config. But that's not really your problem so i think you can merge this and i'll still have a look if we can do something to make the situation better from mx.

PR #637, opened 2016-12-16
Replace System.arraycopy with for-loop

grimmerm:


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/func/LLVMCallNode.java
```diff
@@ -404,6 +393,21 @@ public Object execute(VirtualFrame frame) {
...
+        private static Object[] convertToPrimitiveArgs(Object[] arguments) {
+            CompilerAsserts.compilationConstant(arguments.length);
+            Object[] newArguments = new Object[arguments.length - LLVMCallNode.ARG_START_INDEX];
+            for (int i = LLVMCallNode.ARG_START_INDEX; i < arguments.length; i++) {
+                newArguments[i - LLVMCallNode.ARG_START_INDEX] = arguments[i];
+            }
+            for (int i = 0; i < newArguments.length; i++) {
+                if (newArguments[i] instanceof LLVMAddress) {
+                    newArguments[i] = ((LLVMAddress) newArguments[i]).getVal();
+                }
```
rschatz: It probably doesn't make a difference for performance, but wouldn't this code be more readable if we do this in a single loop?

PR #644, opened 2016-12-21
Parallelize LifetimeAnalysis

jkreindl:


grimmerm:
@jkreindl what is the status of this PR?

jkreindl:
@grimmerm If I remember correctly I considered the code done, but the improvement is debatable. For programs with few and short methods like most of our tests the overhead introduced by parallelization outweighs the gains while there should be an improvement noticable for programs with many and / or large methods. As Sulong seems to be used primarily for implementing other language's C-extensions I do not think the latter case occurs all that often for us, unless we want to run some heavier libraries, so I am unsure whether we should really merge this.

PR #649, opened 2017-01-02
Associate Memory with Type Information

jkreindl:
Can be merged now.


projects/com.oracle.truffle.llvm.types/src/com/oracle/truffle/llvm/types/LLVMAddress.java
```diff
@@ -38,16 +38,42 @@
 
     public static final LLVMAddress NULL_POINTER = fromLong(0);
 
+    private final String symbolName;
+
+    private final long symbolIndex;
```
grimmerm: Can we add the complete type information to LLVMAddress?

jkreindl: We would need to add a dependency on the Parser.API package that contains the Type class for that, but sure. I also intend to add multi-level indexing which would be represented here in the form of a long[].

PR #653, opened 2017-01-04
Implement LLVM IR Writer using Visitor

pointhi:
As discussed with @grimmerm and @mrigger, I try to implement the LLVM IR Writer using a specific Visitor (and not ```toString()``` methodes as done in #628)

I made this PR to discuss implementation details. that means: **Do not merge yet**


projects/com.oracle.truffle.llvm.parser.api/src/com/oracle/truffle/llvm/parser/api/model/symbols/constants/AbstractConstant.java
```diff
@@ -39,6 +40,8 @@ protected AbstractConstant(Type type) {
         this.type = type;
     }
 
+    public abstract void accept(AbstractConstantVisitor visitor);
```
jkreindl: _AbstractConstant_ right now only serves the purpose of storing the type so the individual constants do not have to implement that functionality themselves. The actual top-level interface for constants is _Constant_. I think we should either remove the interface altogether and rename this class to _Constant_ or declare the _accept()_ method in the interface.

jkreindl:
@pointhi I changed the layout of the produced IR a bit to make it a bit more similar to that produced of llvm-dis. I also added an mx option to print the IR to stdout or to a file. For that I merged the llvm.writer project into the parser.bc project.

projects/com.oracle.truffle.llvm.parser.bc/src/com/oracle/truffle/llvm/parser/bc/irwriter/ConstantPrintVisitor.java
```diff
@@ -174,14 +175,12 @@ public void visit(CompareConstant compareConstant) {
 
     @Override
     public void visit(DoubleConstant doubleConstant) {
-        final long bits = Double.doubleToRawLongBits(doubleConstant.getValue());
-        out.print(String.format("0x%x", bits));
+        out.print(String.format(Locale.ROOT, "%e", doubleConstant.getValue()));
```
jkreindl: See http://llvm.org/releases/3.2/docs/LangRef.html#simpleconstants why we should use the Hex-Representation.

PR #655, opened 2017-01-10
Use AbstractConstant instead of toString() for method call

pointhi:


jkreindl:
@pointhi The same is already done in #646. That PR has not been merged yet though.

pointhi:
closing, because same feature was merged in another PR

PR #657, opened 2017-02-06
Colored mx

pointhi:
using ```mx.log``` and ```mx.log_error``` for output, instead of ```print```

This allows us to make coloured text outputs when there is an error

### DIFF
--- a/mx.sulong/mx_sulong.py
+++ b/mx.sulong/mx_sulong.py
@@ -150,7 +150,7 @@ def pullLLVMBinaries(args=None):
     osStr = mx.get_os()
     arch = mx.get_arch()
     if osStr == 'windows':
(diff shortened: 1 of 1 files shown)


mx.sulong/mx_sulong.py
```diff
@@ -150,7 +150,7 @@ def pullLLVMBinaries(args=None):
     osStr = mx.get_os()
     arch = mx.get_arch()
     if osStr == 'windows':
-        print 'windows currently only supported with cygwin!'
+        mx.log_error('windows currently only supported with cygwin!')
```
grimmerm: very nice feature, thanks!

pointhi:
merged via #664

PR #658, opened 2017-02-14
Align the Parser's Variable Naming Scheme with the LLVM Toolset

jkreindl:


jkreindl:
Closed. All changes are also contained in #663.

PR #659, opened 2017-02-15
Add string representation and other improvements to some enums

pointhi:


projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/model/enums/AtomicOrdering.java
```diff
@@ -30,18 +30,25 @@
...
 public enum AtomicOrdering {
-    NOT_ATOMIC(0L),
-    UNORDERED(1L),
-    MONOTONIC(2L),
-    ACQUIRE(3L),
-    RELEASE(4L),
-    ACQUIRE_RELEASE(5L),
-    SEQUENTIALLY_CONSISTENT(6L);
+    NOT_ATOMIC(0L, ""),
+    UNORDERED(1L, "unordered"),
```
grimmerm: please document why these enum values need a name different from the enum value name.
Recent refactorings removed these names because it was not obvious where the names are needed.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/model/enums/BinaryOperator.java
```diff
@@ -31,25 +31,25 @@
...
-    INT_SIGNED_DIVIDE(16),
-    INT_UNSIGNED_REMAINDER(-1),
-    INT_SIGNED_REMAINDER(17),
-    INT_SHIFT_LEFT(-1),
-    INT_LOGICAL_SHIFT_RIGHT(-1),
-    INT_ARITHMETIC_SHIFT_RIGHT(-1),
-    INT_AND(-1),
-    INT_OR(-1),
-    INT_XOR(-1),
+    INT_ADD(13, "add"),
```
grimmerm: same here

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/model/enums/BinaryOperator.java
```diff
@@ -61,12 +61,23 @@ public static BinaryOperator decode(int opcode, boolean isFloatingPoint) {
...
         this.fpmap = fpmap;
+        this.name = name;
     }
 
     private BinaryOperator fp() {
         return fpmap < 0 ? null : values()[fpmap];
     }
+
+    public boolean isFloatingPoint() {
+        return this.ordinal() > INT_XOR.ordinal();
```
grimmerm: This is a public API function that is not used in Sulong. Therefore future refactorings might remove this function.
Please add some JavaDoc/comment why this function exists and what its purpose is.

pointhi: this method is needed to differ between normal and floating point operators in LLVMIR writer.

If this is not sensible to put it into the enum, I can make helper classes for such tasks in my library and this PR would be no longer required at all.

pointhi:
merged via #664

PR #661, opened 2017-02-15
Implement some Visitors for various datatypes

pointhi:


projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/listeners/Types.java
```diff
@@ -217,6 +218,11 @@ public Type get(long index) {
             this.idx = idx;
         }
 
+        @Override
+        public void accept(TypeVisitor visitor) {
+            // TODO: ignore?
```
grimmerm: What's this?

pointhi: because those are private types, which are not exposed. As well as they are only required for resolving while parsing, a visitor implementation of them is not useful for most use cases.

Furthermore those classes are private, so we would need to make them public if we are implementing visitors for them.

pointhi:
merged via #664

PR #667, opened 2017-03-10
Change some toString to getIrString method, to be clear about it's behaviour

pointhi:
```toString()``` should not be used to output the LLVM IR name of the entry, so I changed it to the ```getIrString``` method which is also used for the other enums


pointhi:
Fixed

PR #668, opened 2017-03-13
Fix su-pulldragonegg to either use LLVM-3.2 or LLVM-3.3

pointhi:
Otherwise, the command would fail when there is LLVM-3.8 or greater
installed on the system


pointhi:
Fixed

PR #678, opened 2017-04-03
Add improved support of opaque type

pointhi:
Actually, the opaque type has a name attached like a struct. This patch adds support for that


projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/types/OpaqueType.java
```diff
@@ -75,4 +75,29 @@ public String toString() {
...
+        int result = 1;
+        result = prime * result + ((name == null) ? 0 : name.hashCode());
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
```
okutane: null check is included in instanceof check, you can remove this 2 lines. :)

pointhi: well, that happens when someone uses the eclipse codegenerator, without further checking :)

PR #683, opened 2017-04-10
Fix Running Testsuites with LLVM 3.9 and 4.0

jkreindl:


jkreindl:
closed because of erroneous force push without fetching first

Issue #654, opened 2017-01-10
Sulong with Swift?


ambientlight:
Greeting guys,

Has anybody tried (or started) integrating Sulong with swift toolchain? Or its too early yet?
It just feels quite promising that sulong can potentially allow swift to target JVM. 
(I just recently discovered sulong).

grimmerm:
Hi,

Swift would be interesting, yes. However, we did not yet try it. Currently we test Sulong with C/C++ and Fortran compiled with Clang 3.2 or GCC 4.6. 

-Matthias

grimmerm:
I did not look at the LLVM IR produced by Swift yet. So I can not answer this question; sounds like a nice experiment though ;-) If you start playing around with Sulong and Swift, I'm happy to answer any questions you might have: matthias.grimmer@oracle.com or Skype: grimmer_m

ambientlight:
Great, thanks!

RUSshy:
Hi, i find hard to see benefit of this
Swift produce native, self contained binary, why would someone add a huge dependency (jre)?
For mobile development it is a big no because of the size
Or i'm understanding wrong? Does GraamVM compiler produce self contained executable?
Is there a place where i can read more about this project ?

hackyourlife:
> Hi, i find hard to see benefit of this
> Swift produce native, self contained binary, why would someone add a huge dependency (jre)?

There is exactly one reason why you might want to use Sulong: to get seamless interoperability with other Truffle languages like Ruby, JavaScript, R or Python. Otherwise you could just compile the LLVM IR to native code and completely forget Sulong (and most of the time that would also be faster).

> Or i'm understanding wrong? Does GraamVM compiler produce self contained executable?

Depends. The Graal compiler can produce a self contained executable (= compile Sulong to a self contained executable; you can search for something called "SubstrateVM"). It can also run on top of a GraalVM which is just a JVM + Graal compiler as JIT. In both cases you end up with something like the `lli` command from LLVM, so you have to pass the bitcode to Sulong at runtime and Sulong will execute it.

> Is there a place where i can read more about this project ?

There are papers about both Graal/Truffle as well as Sulong. See [docs/PUBLICATIONS.md](https://github.com/graalvm/sulong/blob/master/docs/PUBLICATIONS.md) and [Graal/Publications and Presentations](https://wiki.openjdk.java.net/display/Graal/Publications+and+Presentations).

PR #686, opened 2017-04-22
Use binary version identifiers instead of textual

okutane:
Hello!
I'm proposing a small step towards generalization of sulong so it can be used with more LLVM bitcode producers.
Originally, I've tried to use it with apple clang (which version string is different from original clang's) and it's failed to parse, while bitcode file was totally compatible.

So I've looked into http://llvm.org/docs/BitCodeFormat.html#module-code-version-record and I think version number from MODULE_BLOCK is more strict and reliable than version strings.

This change is also needed for stuff like #654 

I'm actually using sulong just for parsing, but I want to parse as many llvm based languages as possible, so more improvements toward generalization are to come (if you don't mind :) )

Changes overview:
- sulong.LLVM option is removed, as we now using correct version taken from bitcode file being parsed;
- version specific parsers are renamed to match bitcode format versions not clang versions;
- listeners.Identification now does nothing, but kept to prevent warning about ignored records.

### DIFF
--- a/mx.sulong/mx_testsuites.py
+++ b/mx.sulong/mx_testsuites.py
@@ -77,10 +77,10 @@ def run(vmArgs, unittest, versionFolder, extraOption=None):
         return mx_unittest.unittest(command)
 
 def run32(vmArgs, unittest, extraOption=None):
-    run(vmArgs + ['-Dsulong.LLVM=3.2'], unittest, 'v32', extraOption)
+    run(vmArgs, unittest, 'v32', extraOption)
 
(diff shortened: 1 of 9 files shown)


grimmerm:
Dmitry, thanks for this contributions (and all future ones; we are looking forward to them)!
Please find a few comments in the source code.

okutane:
Hm.. my implementation is totally incorrect. I've misinterpret the point of 3.2 / 3.8 versioning. And my fix should be not about changing switch logic, but about generalizing these versions. I think I'll create different PR for that later.

PR #687, opened 2017-04-28
Fix resume instruction implementation

pointhi:


jkreindl:
Fixed in #685

jkreindl:
Yes, you are right. Looks like I remembered the content of your PR wrong, I just thought of the changes in Function.java.

PR #692, opened 2017-05-08
Check whether a File is available before creating a SourceSection

jkreindl:
Debug Information might reference files that are not present at runtime but were present at compiletime.


jkreindl:
Obsoleted by #694

PR #694, opened 2017-05-11
Reimplement SourceSection Generation from Debug Information in Bitcode Files

jkreindl:
We try to include every bit of information we can get from Metadata into SourceSections but still provide sensible defaults in case sources are not accessible.


jkreindl:
All changes contained in #696

PR #695, opened 2017-05-15
Pass Correct Arguments to LLVMCopySign

jkreindl:


jkreindl:
All changes contained in #696

PR #697, opened 2017-05-16
Better Encapsulation of Optional Debug Information

jkreindl:
A LLVMFunctionStartNode is always guaranteed a SourceSection, but other Debug Information need not be present. This encapsulates it into an object that is only created if really necessary, instead of always with mostly empty content.


jkreindl:
Closed because of misinterpretation of the problem.

PR #698, opened 2017-05-18
Add mx option to compile tests with Debug Information

jkreindl:
Introduce the SULONG_COMPILE_WITH_DEBUGINFO environment variable. If set to 'TRUE', mx will compile the tests with the '-g' option to include debug information into the bitcode files.

We cannot use this as default since Truffle Source objects cannot correctly handle some file formats.
An example of this would be 'gcc-5.2.0/gcc/testsuite/gcc.dg/cpp/mac-eol-at-eof.c'. It uses only single '\r' characters as line endings.


jkreindl:
Included in #704.

PR #704, opened 2017-05-26
Improve Debugging Support

jkreindl:
DO NOT MERGE YET!

- Provide the 'StatementTag' and 'CallTag' annotations to enable instrumentation support, e.g. allow for stepping through the original C source code with the [Truffle Debugger for Netbeans](http://wiki.apidesign.org/wiki/Truffle#Debugger_in_NetBeans_8.1). This will require further work though.

- Reduce Parser Log Spam if 'sulong.Debug' is set.

- Introduce the 'sulong.TraceSourceExecution' option to trace Sulong's execution of C source code.

- Introduce the SULONG_COMPILE_WITH_DEBUGINFO environment variable. If set to 'TRUE', mx will compile the tests with the '-g' option to include debug information into the bitcode files. We cannot just use this as default since Truffle Source objects cannot correctly handle some file formats.
An example of this would be 'gcc-5.2.0/gcc/testsuite/gcc.dg/cpp/mac-eol-at-eof.c'. It uses only single '\r' characters as line endings.


mx.sulong/mx_tools.py
```diff
@@ -102,7 +102,12 @@ def getTool(self, inputFile):
 
     def run(self, inputFile, outputFile, flags):
         tool = self.getTool(inputFile)
-        return self.runTool([mx_sulong.findLLVMProgram(tool, ['3.2']), '-c', '-emit-llvm', '-o', outputFile] + flags + [inputFile], errorMsg='Cannot compile %s with %s' % (inputFile, tool))
+        cmd = [mx_sulong.findLLVMProgram(tool, ['3.2'])]
+        if os.environ.get('SULONG_COMPILE_WITH_DEBUGINFO') in ['True', 'TRUE', 'true']:
```
mrigger: `.lower()` == 'true' would also catch `TrUe` and `tRUE`. ;)

PR #710, opened 2017-06-02
	Fix Parsing Forward Referenced Symbols

jkreindl:
LLVM Bitcode often contains type references for forward referenced symbols. These need to be parsed correctly.


grimmerm:
Will be merged after a rebase

PR #712, opened 2017-06-20
add parameterattribute support to call and invoke, and fix byvalue handling of calls

pointhi:
* Parser Parameter Attributes inside calls and invokes
* add interface to allow general access for those information
  @grimmerm: please check if this interface declaration doesn't interfere with design decisions regarding call and invoke.


pointhi:
fixed: https://github.com/graalvm/sulong/commit/9675b20f1bfc218d16ccfae383050b287e067314

PR #714, opened 2017-06-22
Implement Preliminary Instrumentation Support

jkreindl:
The wrapper nodes for instrumentation should not have any fields but the delegate and probe nodes. For this we need to refactor some node implementations.


grimmerm:
LGTM
@christianhaeubl, can you spot any obvious performance issues?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMBrUnconditionalNode.java
```diff
@@ -29,33 +29,57 @@
...
+
+        @Override
+        public int getSuccessor() {
+            return successor;
+        }
 
-    public int getSuccessor() {
-        return successor;
+        @Override
+        public void execute(VirtualFrame frame) {
```
grimmerm: Should we somehow assert that this execute is never called?
`CompilerDirectives.transferToInterpreter(); throw new ISE();`

grimmerm: Or is this method actually called?
At least, add some clarification whether this is called or not.

grimmerm: Ah, I just saw that you actually calling this from the dispatch node.

grimmerm: FWDing @christianhaeubl comment: 
would be better to put the comment into the implementation (nothing todo but we need ...) - do all LLVMControlFlowNodes now need an execute method? if so, then:
- the class hierarchy could be unified a bit (currently some have the executeCondition method and some don't have anything)
- what about other classes that are annotated with @Instrumentable but missing an execute method? (at least LLVMRetNode and LLVMResumeNode)

PR #689, opened 2017-05-04
Initial commit to add support for the Rust language

anatol1234:
Adds support for running simple Rust programs.
Example `fib.rs`:
```rust
const  X: u64 = 30;

fn fib(n: u64) -> u64 {
    if n == 0{
       	0
    }else if n == 1 {
	1
    }else{
	fib(n - 1) + fib(n - 2)
    }
}

fn main(){
	println!("Fibonacci of {}: {}", X, fib(X))
}
```
Produce bitcode file with: `rustc --emit=llvm-bc fib.rs`
Run with: `mx lli -Dpolyglot.llvm.libraries=<Rust libstd> fib.bc`

PR #726, opened 2017-07-18
Implement some of the LLVM Arithmetic builtins

pointhi:
Implement some builtin which are used for arithmetic calculation including overflow checks.

There was also a implementation error with the ```"@llvm.uadd.with.overflow.i32``` builtin, which should be fixed now.

Documentation:
* https://clang.llvm.org/docs/LanguageExtensions.html#multiprecision-arithmetic-builtins
* https://llvm.org/docs/LangRef.html#llvm-sadd-with-overflow-intrinsics
* https://github.com/llvm-mirror/clang/blob/release_40/include/clang/Basic/Builtins.def#L1303

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/arith/LLVMSAddWithOverflow.java
+++ b/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/arith/LLVMSAddWithOverflow.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2017, Oracle and/or its affiliates.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are
+ * permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list of
+ * conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this list of
+ * conditions and the following disclaimer in the documentation and/or other materials provided
+ * with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may be used to
+ * endorse or promote products derived from this software without specific prior written
+ * permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS
+ * OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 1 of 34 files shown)


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/arith/LLVMSAddWithOverflow.java
```diff
@@ -0,0 +1,102 @@
...
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMExpressionNode;
+
+public class LLVMSAddWithOverflow {
+
+    @NodeChildren({@NodeChild(value = "left", type = LLVMExpressionNode.class), @NodeChild(value = "right", type = LLVMExpressionNode.class),
+                    @NodeChild(value = "target", type = LLVMExpressionNode.class)})
+    public abstract static class LLVMSAddWithOverflowI8 extends LLVMBuiltin {
+
+        @Specialization
+        public LLVMAddress executeIntrinsic(byte left, byte right, LLVMAddress addr) {
```
grimmerm: where is the memory (addr) allocated?
Is there bitcode for that (e.g. stack allocation)?

pointhi: I used the old uadd node as implementation reference. At least in llvm ir there is no allocation of the memory used in this function visible.

anatol1234:
PR #689 already contains implementations for the I16, I32 and I64 overloads of these builtins. We should coordinate with respect to these builtins. Do you know any constraints which guarantee that the results of these builtins are aligned the way you write them into LLVM memory? If this is not the case, we will have to pay attention to the specified data layout.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/arith/LLVMUAddWithOverflow.java
```diff
@@ -41,16 +41,60 @@
...
+
+    @NodeChildren({@NodeChild(value = "left", type = LLVMExpressionNode.class), @NodeChild(value = "right", type = LLVMExpressionNode.class),
+                    @NodeChild(value = "target", type = LLVMExpressionNode.class)})
     public abstract static class LLVMUAddWithOverflowI32 extends LLVMBuiltin {
 
-        private static final int OVERFLOW_FIELD_OFFSET = 4;
+        @Specialization
+        public LLVMAddress executeIntrinsic(int left, int right, LLVMAddress addr) {
+            final int res = left + right;
+            final boolean overflow = (((~res & left) | (~res & right) | (left & right)) & (1 << (Integer.SIZE - 1))) != 0;
```
anatol1234: Is there a particular reason, why you write "& (1 << (Integer.SIZE - 1))) != 0" instead of "< 0"? I used "< 0", because I thought it should work like for the signed version.

pointhi: thx, I changed it to the simpler version using less than

pointhi:
@anatol1234 Can you verify the new implementation, if it also works with rust? https://github.com/graalvm/sulong/pull/726/commits/f8841382e36a5a0fb39e2eae9d49463f7354fe94

PR #729, opened 2017-07-18
Enable Local Variable Inspection for Debuggers

jkreindl:
Should this feature be enabled Sulong will maintain an additional value on the stack that holds representations of the source-level state of the local and global variables of the running program.


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugObjectImpl.java
```diff
@@ -0,0 +1,239 @@
...
+                case Long.SIZE:
+                    return value.readLongUnsigned(offset);
+
+                default:
+                    return value.readUnknown(offset, type.getSize());
+            }
+        }
+    }
+
+    static final class Complex extends LLVMDebugObjectImpl {
```
grimmerm: It's rather "Structured"

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMAddressValueProvider.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    @Override
+    public boolean canRead() {
+        return !LLVMAddress.nullPointer().equals(baseAddress);
+    }
+
+    @Override
+    public boolean readBoolean(long offset) {
+        if (!canRead()) {
```
grimmerm: Add a CompilerDirectives.transferToInterpreter() here (and all other places where you are using this pattern).

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
```diff
@@ -0,0 +1,130 @@
...
+import com.oracle.truffle.llvm.runtime.global.LLVMGlobalVariableAccess;
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMExpressionNode;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@NodeChild(value = "accessor", type = LLVMExpressionNode.class)
+public abstract class LLVMDebugDeclaration extends LLVMExpressionNode {
+
+    private final FrameSlot debugSlot;
```
grimmerm: What's stored in this debugSlot?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
```diff
@@ -0,0 +1,130 @@
...
+    public Object executeGeneric(VirtualFrame frame) {
+        return null;
+    }
+
+    @Specialization
+    public Object readAddress(VirtualFrame frame, LLVMAddress address) {
+        final Object debugSlotVal = frame.getValue(debugSlot);
+        DynamicObject debugObj;
+
+        if (debugSlotVal instanceof DynamicObject) {
```
grimmerm: that looks like some lazy initialization of debugObj. Can we avoid this check (which the compiler most likely won't be able to remove)? Let's discuss that.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
```diff
@@ -0,0 +1,130 @@
...
+
+        debugObj = LLVMDebugSlotType.findOrAddGlobalsContainer(debugObj);
+
+        final LLVMAddress address = globalAccess.getNativeLocation(global);
+        final LLVMDebugObject object = instantiate(varType, 0L, new LLVMAddressValueProvider(address));
+        debugObj.define(varName, object);
+        return null;
+    }
+
+    private LLVMDebugObject instantiate(LLVMDebugType type, long baseOffset, LLVMDebugValueProvider value) {
```
grimmerm: we might want to hide that behind a TruffleBoundary

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugObjectImpl.java
```diff
@@ -0,0 +1,239 @@
...
+            return NO_KEYS;
+        }
+
+        @Override
+        public Object getMember(Object identifier) {
+            return null;
+        }
+
+        @Override
+        public Object getValue() {
```
grimmerm: we might want to hide getValue (and probably all other methods of LLVMDebugObjectImpl) behind TruffleBoundaries

projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/debug/LLVMDebugStructLikeType.java
```diff
@@ -0,0 +1,118 @@
...
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.runtime.debug;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+
+public final class LLVMDebugStructLikeType extends LLVMDebugType {
```
grimmerm: All these types are used at runtime - we need to ensure that the code of these types does not end up in compiled code (annotate with @TruffleBoundary what should not be inlined by PE, hide dedicated slow-path code (e.g. error handling) behind a transferToInterpreter, and assert those parts where you know that they will never end up in compiled code with a neverPartOfCompilation).
A defense programming style here is probably a good idea and we can save time when hunting performance issues.

projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/debug/LLVMDebugType.java
```diff
@@ -0,0 +1,114 @@
...
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.runtime.debug;
+
+import java.util.function.Supplier;
+
+public abstract class LLVMDebugType {
```
grimmerm: see comment above; this applies to all runtime data that you've introduced.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMAddressValueProvider.java
```diff
@@ -0,0 +1,274 @@
...
+        this.baseAddress = baseAddress;
+    }
+
+    @Override
+    public boolean canRead() {
+        return !LLVMAddress.nullPointer().equals(baseAddress);
+    }
+
+    @Override
+    @TruffleBoundary
```
grimmerm: Why are you adding a TruffleBoundary here? that seems fine for compilation

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMAddressValueProvider.java
```diff
@@ -0,0 +1,274 @@
...
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
+import com.oracle.truffle.llvm.runtime.LLVMAddress;
+import com.oracle.truffle.llvm.runtime.debug.LLVMDebugValueProvider;
+import com.oracle.truffle.llvm.runtime.floating.LLVM80BitFloat;
+import com.oracle.truffle.llvm.runtime.memory.LLVMMemory;
+
+import java.math.BigInteger;
+
+final class LLVMAddressValueProvider implements LLVMDebugValueProvider {
```
grimmerm: I'm not sure if it is really necessary to wrap the `baseAddress` into an `LLVMAddressValueProvider` (after all, you are not adding any additional members).
Maybe this class could be a simple util class with static functions (e.g. `public static boolean readBoolean(LLVMAddress base, long offset)` )

projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/debug/LLVMDebugValueProvider.java
```diff
@@ -0,0 +1,78 @@
...
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.runtime.debug;
+
+import com.oracle.truffle.llvm.runtime.LLVMAddress;
+import com.oracle.truffle.llvm.runtime.floating.LLVM80BitFloat;
+
+import java.math.BigInteger;
+
+public interface LLVMDebugValueProvider {
```
grimmerm: There is only one class implementing this interface - is that interface really necessary?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
```diff
@@ -0,0 +1,133 @@
...
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMExpressionNode;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@NodeChild(value = "accessor", type = LLVMExpressionNode.class)
+public abstract class LLVMDebugDeclaration extends LLVMExpressionNode {
+
+    /** Contains the object that stores the source-level variables. */
+    private final FrameSlot sourceValuesContainerSlot;
```
grimmerm: Is it correct that if the bc file was compiled without debug information, this FrameSlot is never added to the frame and sourceValuesContainerSlot is null?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
```diff
@@ -0,0 +1,133 @@
...
+    public Object executeGeneric(VirtualFrame frame) {
+        return null;
+    }
+
+    @Specialization
+    public Object readAddress(VirtualFrame frame, LLVMAddress address) {
+        final Object debugSlotVal = frame.getValue(sourceValuesContainerSlot);
+        DynamicObject debugObj;
+
+        if (debugSlotVal instanceof DynamicObject) {
```
grimmerm: That should be done using specializations

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
```diff
@@ -0,0 +1,133 @@
...
+        debugObj.define(varName, object);
+        return null;
+    }
+
+    @Specialization
+    public Object readGlobal(VirtualFrame frame, LLVMGlobalVariable global, @Cached("createGlobalAccess()") LLVMGlobalVariableAccess globalAccess) {
+        final Object debugSlotVal = frame.getValue(sourceValuesContainerSlot);
+        DynamicObject debugObj;
+
+        if (debugSlotVal instanceof DynamicObject) {
```
grimmerm: Same here

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugObjectImpl.java
```diff
@@ -0,0 +1,373 @@
...
+import com.oracle.truffle.llvm.runtime.debug.LLVMDebugDecoratorType;
+import com.oracle.truffle.llvm.runtime.debug.LLVMDebugObject;
+import com.oracle.truffle.llvm.runtime.debug.LLVMDebugType;
+import com.oracle.truffle.llvm.runtime.debug.LLVMDebugValueProvider;
+import com.oracle.truffle.llvm.runtime.floating.LLVM80BitFloat;
+
+import java.util.Map;
+import java.util.Objects;
+
+abstract class LLVMDebugObjectImpl implements LLVMDebugObject {
```
grimmerm: Having both, LLVMDebugObject and LLVMDebugObjectImpl does not make sense.
You only ever implement LLVMDebugObject here.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMAddressValueProvider.java
jkreindl: The ValueProvider interface allows us to use different data providers with the same object class. At the moment the LLVMAddress is the only backing data structure we support, but this might change in the future, e.g. when I implement the @llvm.debug.value intrinsic or when we add debugging support to SafeSulong.

projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/debug/LLVMDebugValueProvider.java
jkreindl: See my comment to LLVMAddressValueProvider.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
jkreindl: If the bc file was compiled without debug information it will not contain any calls @llvm.debug.declaration intrinsic which could be resolved to an LLVMDebugDeclaration, so the node would not be created in the first place. Also, if Sulong is not executed with -Dpolyglot.llvm.enableLVI=true these intrinsics are just ignored and the FrameSlot is not added in the first place.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugObjectImpl.java
jkreindl: We need access to LLVMDebugObject in Sulong.java::findMetaObject(LLVMContext, Object) since this is how the debugger gets the value to display in the type field. I think the implementation class makes more sense in the llvm.nodes package but this can of course be changed.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugDeclaration.java
jkreindl: This would require an additional node, but it would be much cleaner. I will change that.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/llvm/debug/LLVMDebugObjectImpl.java
grimmerm: OK, makes sense. Thanks for the clarification.

PR #730, opened 2017-07-19
Fix Parsing of Linkage

jkreindl:
There are some new constants in LLVM 4.0.


rschatz:
Looks weird that there are two constants that mean the same thing. Where did you get these additional constants from? Is this documented somewhere on the LLVM side?
I see no mention of these additional constants in http://releases.llvm.org/4.0.0/docs/BitCodeFormat.html#linkage-type

I can reproduce the problem locally, i.e., I also see the new constants in bitcode files produced by clang 4.0.0, but I'd like to understand where they come from...

jkreindl:
@rschatz I got those constants from the LLVM 4.0 source code (see [here](https://llvm.org/doxygen/BitcodeReader_8cpp_source.html#l00821)). Apparently LLVM introduced a new format for storing linkage types with additional data in the bitcode files which lead them to introduce new constants. They still support the old constants too to remain backwards compatible.

rschatz:
Ok, so their documentation is just outdated. Thanks for researching this ;)
I'll go ahead and merge this.

PR #732, opened 2017-07-23
Add support for "atomicrmw" instruction

Godin:
During experiments with Sulong at [JCrete](http://www.jcrete.org/) with @jtulach and @alexsnaps we found that [`atomicrmw` instruction](https://llvm.org/docs/LangRef.html#atomicrmw-instruction) was not implemented.

---

Despite the fact that @graalvmbot complains about OCA, at http://www.oracle.com/technetwork/community/oca-486395.html#m you can find that it was signed.


rschatz:
The @graalvmbot has a manually maintained whitelist, and it only scans open PRs, not already merged ones. We added you to the whitelist for the future. Sorry for the confusion, and thanks again for the contributions ;)

dougxc:
The underlying problem is that the "database" of OCA signatories is a Word document somewhere which makes it rather hard for graalvmbot to automatically stay in sync with it.

PR #733, opened 2017-07-23
Add "llvm.assume" builtin

Godin:
During experiments with Sulong at [JCrete](http://www.jcrete.org/) with @jtulach and @alexsnaps we found that [`llvm.assume` builtin](https://llvm.org/docs/LangRef.html#llvm-assume-intrinsic) was not implemented.

---

Despite the fact that @graalvmbot complains about OCA, at http://www.oracle.com/technetwork/community/oca-486395.html#m you can find that it was signed.


Godin:
@dougxc thank you for the merge. I'm wondering if possible to adjust @graalvmbot so that to prevent future complains about OCA?

grimmerm:
Thanks!
This is was an automerge - I'll put you on the whitelist.

Godin:
@grimmerm thanks! :thumbsup:

PR #736, opened 2017-07-26
Implement Zero Extend Nodes for variable integers

pointhi:


rschatz:
The change itself looks good. Can you add a few small unit tests to the SulongSuite that execute the new code?

PR #741, opened 2017-08-10
Add support for Rust version 1.19.0

anatol1234:
The signature for the std::rt::lang_start function has changed with version 1.19.0. This pull request includes support for the new signature and fixes some other minor issues that were revealed when running bitcode produced by the latest version of the Rust compiler.

### DIFF
--- a/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/listeners/Function.java
+++ b/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/listeners/Function.java
@@ -763,13 +763,13 @@ public final class Function implements ParserListener {
         int i = 0;
         int vector = getIndex(args[i++]);
 
-        Type type;
+        Type vectorType;
         if (function.isValueForwardRef(vector)) {
-            type = types.get(i++);
+            vectorType = types.get(args[i++]);
         } else {
-            type = ((VectorType) function.getValueType(vector)).getElementType();
+            vectorType = function.getValueType(vector);
         }
-
+        Type type = ((VectorType) vectorType).getElementType();
         int index = getIndex(args[i]);
 
         emit(ExtractElementInstruction.fromSymbols(function.getSymbols(), type, vector, index));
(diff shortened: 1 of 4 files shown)


projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/listeners/Function.java
```diff
@@ -982,6 +982,20 @@ private static int getAlign(long argument) {
         return (int) argument & (Long.SIZE - 1);
     }
 
+    private static Type getExtractedElementType(Type type) {
+        Type valType = type;
+        if (valType instanceof PointerType) {
```
jkreindl: According to [the llvm documentation](http://releases.llvm.org/4.0.1/docs/LangRef.html#extractelement-instruction) and also looking at the llvm 4.0.1 source code this should only ever be a vector type to be valid bitcode. Did you actually see these in your bitcode files? In any case, these are all subclasses of AggregateType which declares the 'getElementType(int)' method. Could you change this to `if (valType instanceof AggregateType) val = ((AggregateType) val).getElementType(0)`.

anatol1234: Thank you for pointing that out. I was missing that I am getting completely wrong types from `types.get(i++);` in the case of a forward reference. `types.get(args[i++]);` retrieves the correct type for me which is always a vector type as specified in the docs. I will change my fix accordingly. However, I am wondering if there exist cases where `types.get(i++);` would retrieve the correct type rather than `types.get(args[i++]);`.

jkreindl: I completely overlooked the `types.get(i++)`, thanks for noticing! `i` is only the index in the argument array and will only ever retrieve the correct type if it happens to be equal to `args[i]`. This same mistake happens in several other places too.

anatol1234: I have already wondered if `types.get(i++)` is wrong in all places where it appears. I just couldn't imagine that could have gone unnoticed. Will you fix that for the other instructions?

jkreindl: I fixed this in #742. Forward referenced symbols are not the common case and the instructions this error appeared in are not the most common either, so it is entirely possible for this to go unnoticed.

rschatz:
LGTM, thanks.

PR #742, opened 2017-08-11
Fix Parsing of Forward Referenced Types

jkreindl:
Sulong was parsing some forward referenced types incorrectly. Thanks @anatol1234 for making me aware of this.


rschatz:
LGTM, thanks!

PR #754, opened 2017-08-24
Implement vectorised byte swap intrinsics.

mikehearn:
This fixes shasum from coreutils.

Note, I have signed and sent off the OCA today. I will hopefully appear on the list in a day or two. I figured I could as well get code review started early.


mikehearn:
OK, this should be good to merge now.

mrigger:
Thanks and sorry for only seeing now that the change is ready to merge. I have pushed the change to our internal gate. Are there other PRs that I should push?

PR #755, opened 2017-08-24
Allow function selects with function handles too.

mikehearn:
Note, I have signed and sent off the OCA today. I will hopefully appear on the list in a day or two. I figured I could as well get code review started early.


grimmerm:
@mikehearn, we refactored how we handle functions - an LLVMFunctionHandle is now an LLVMAddress. does this make this PR obsolete?

mikehearn:
It might well do. I'll have to find some more sulong time and see if the apps I was testing work without this.

rschatz:
I'm pretty sure this is obsolete after the function handle refactoring. Closing.
Thanks anyway for the effort.

PR #756, opened 2017-08-24
Support all bit widths in switch instructions.

mikehearn:
I took a crack at solving this, but as you can see from the // XXX comment, I was unsure of how to handle UnexpectedResultException at one point. Advice appreciated!


jkreindl:
LLVMIVarbit does not implement equals()/hashcode() so the check at LLVMDispatchBasicBlockNode.java::115 will fail to find the correct case value.

grimmerm:
Thanks Mike, I'm closing this PR because this seems to be fixed already.

Issue #748, opened 2017-08-22
Forward reference types still not fully supported


mikehearn:
I saw a PR go in recently that said it added support for forward reference types, but the support seems incomplete.

It is possible to have bitcode that defines structs before all the member types are defined. Reading the LLVM format docs, it seems the parser is expected to use a type table. Unfortunately Sulong has that table but resolves indexes to objects as the records are read, which is too early. There's some infrastructure to try and tackle this for a few types, but it doesn't work for structs.

The fix is probably to load structs whilst parsing as an intermediate type that keeps around the indexes, then re-iterate over the type table and convert to StructureType after the parse is fully completed.

grimmerm:
@jkreindl can you have a look pls

Issue #747, opened 2017-08-20
Bitcode for fish shell fails to parse


mikehearn:
After building fish at commit ba53242b266a0901b176ef7f7de872dc6b5e9956 with wllvm, I get this when trying to execute it:

```
org.graalvm.polyglot.PolyglotException: java.lang.RuntimeException: java.io.IOException: Error while trying to parse /Users/mike/Source/graal/sulong/fish-shell/fish.bc
	at com.oracle.truffle.llvm.parser.listeners.Function.getElementPointerType(Function.java:1010)
	at com.oracle.truffle.llvm.parser.listeners.Function.createGetElementPointer(Function.java:810)
	at com.oracle.truffle.llvm.parser.listeners.Function.record(Function.java:267)
	at com.oracle.truffle.llvm.parser.scanner.LLVMScanner.passRecordToParser(LLVMScanner.java:343)
	at com.oracle.truffle.llvm.parser.scanner.LLVMScanner.abbreviatedRecord(LLVMScanner.java:164)
	at com.oracle.truffle.llvm.parser.scanner.LLVMScanner.scanNext(LLVMScanner.java:157)
	at com.oracle.truffle.llvm.parser.scanner.LLVMScanner.parse(LLVMScanner.java:96)
	at com.oracle.truffle.llvm.parser.BitcodeParserResult.getFromSource(BitcodeParserResult.java:66)
	at com.oracle.truffle.llvm.parser.LLVMParserRuntime.parse(LLVMParserRuntime.java:89)
	at com.oracle.truffle.llvm.Runner.parseBitcodeFile(Runner.java:179)
	at com.oracle.truffle.llvm.Runner.parse(Runner.java:69)
	at com.oracle.truffle.llvm.Sulong.parse(Sulong.java:79)
	at com.oracle.truffle.api.TruffleLanguage$ParsingRequest.parse(TruffleLanguage.java:489)
	at com.oracle.truffle.api.TruffleLanguage.parse(TruffleLanguage.java:773)
	at com.oracle.truffle.api.TruffleLanguage$LanguageImpl.parse(TruffleLanguage.java:1443)
	at com.oracle.truffle.api.vm.PolyglotContextImpl.eval(PolyglotContextImpl.java:530)
	at org.graalvm.polyglot.Context.eval(Context.java:145)
	at com.oracle.truffle.llvm.Sulong.executeMain(Sulong.java:133)
	at com.oracle.truffle.llvm.Sulong.main(Sulong.java:117)
[...]

PR #757, opened 2017-08-24
Refactor Type Parsing to handle Forward-Referenced Types

jkreindl:
Addresses #747 and #748. Forward references in the type table were barely supported until now.

### DIFF
--- a/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/types/ArrayType.java
+++ b/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/types/ArrayType.java
@@ -29,16 +29,23 @@
  */
 package com.oracle.truffle.llvm.runtime.types;
 
+import com.oracle.truffle.api.CompilerAsserts;
+import com.oracle.truffle.api.CompilerDirectives.CompilationFinal;
 import com.oracle.truffle.llvm.runtime.types.visitors.TypeVisitor;
 
 public final class ArrayType extends AggregateType {
 
-    private final Type elementType;
+    @CompilationFinal private Type elementType;
     private final int length;
 
     public ArrayType(Type type, int length) {
         this.elementType = type;
         this.length = length;
+    }
+
+    public void setElementType(Type elementType) {
+        CompilerAsserts.neverPartOfCompilation();
+        this.elementType = elementType;
     }
 
     @Override
@@ -125,5 +132,4 @@ public final class ArrayType extends AggregateType {
         }
         return true;
     }
-
 }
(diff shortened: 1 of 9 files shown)


projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/types/FunctionType.java
```diff
@@ -55,6 +57,11 @@ public Type getReturnType() {
         return returnType;
     }
 
+    public void setReturnType(Type returnType) {
+        CompilerAsserts.neverPartOfCompilation();
```
grimmerm: You have to invalidate all compilation units that depend on this field (similar to what we do in PointerType).
If possible, make returnType final. 
If not, you have to use an assumption (see PointerType) to make this save.

projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/types/ArrayType.java
```diff
@@ -29,18 +29,25 @@
...
 package com.oracle.truffle.llvm.runtime.types;
 
+import com.oracle.truffle.api.CompilerAsserts;
+import com.oracle.truffle.api.CompilerDirectives.CompilationFinal;
 import com.oracle.truffle.llvm.runtime.types.visitors.TypeVisitor;
 
 public final class ArrayType extends AggregateType {
 
-    private final Type elementType;
+    @CompilationFinal private Type elementType;
```
grimmerm: Se comment below.

projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/types/FunctionType.java
jkreindl: This is fixed in the next commit.

mikehearn:
Interesting approach! Can it be documented somewhere why it's using assumptions in this way? Is it because Sulong wants to parse types lazily? I'd thought the fix would be restricted to the parsing code and wouldn't affect execution or compilation.

jkreindl:
The type objects are not private to the parser, they are also used at runtime. I needed to make some fields non-final to replace their values when a forward reference is resolved. I instead annotated them as `@CompilationFinal` which directs the compiler to still treat the values as constant and perform the optimizations this enables. The assumption ensures that a change in this field, wherever it happens, is propagated correctly to all instances of compiled code.

rschatz:
To clarify a bit further: These assumptions are probably *never* invalidated at runtime, since we don't expect the setter to be called after compilation. This is just defensive programming. If we wouldn't have the assumption, and someone calls the setter after code referencing the getter is compiled, this would produce an almost impossible to find bug, since the updated value would just be randomly used or not used, depending on when exactly the method using it was compiled.

On the other hand, with the assumption, the same bug would result in a deoptimization, which might be a performance problem, but it's still correct code. This is relatively easy to diagnose, e.g. by tracing deoptimizations and recompilations.

Also, as long as assumptions are not invalidated, they are almost free (just a little bit of memory, and no runtime cost at all). The "if" checking the assumption will be removed completely from the compiled code, there will be no runtime check left.

mikehearn:
Yep, makes sense. I was thinking that if the code was being lazily parsed, it might end up with lots of deopts as the type info is filled out, but maybe not.

PR #763, opened 2017-08-31
Fix padding error in StructureType.

rmosaner:
The padding for a structure element should be calculated before the element's size is added to the offset.


projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/types/StructureType.java
```diff
@@ -124,10 +124,10 @@ public int getOffsetOf(int index, DataSpecConverter targetDataLayout) {
         int offset = 0;
         for (int i = 0; i < index; i++) {
             final Type elementType = types[i];
-            offset += elementType.getSize(targetDataLayout);
             if (!isPacked) {
                 offset += Type.getPadding(offset, elementType, targetDataLayout);
             }
+            offset += elementType.getSize(targetDataLayout);
```
grimmerm: very good catch! Thanks!

grimmerm:
LGTM

PR #753, opened 2017-08-24
Support atexit taking function descriptors as well as function handles.

mikehearn:
Correct the atexit return type and add some documentation.

Note, I have signed and sent off the OCA today. I will hopefully appear on the list in a day or two. I figured I could as well get code review started early.

PR #774, opened 2017-09-13
x86_64 Syscall implementation in Java

pekd:
This PR adds partial support for file I/O via syscalls (`open`/`close`/`read`/`write`/`readv`/`writev`). A few additional syscalls (e.g. `exit`, `clock_gettime`, `uname`, …) are (partially) implemented.

The bitcode entry point is changed from `main` to `_start` and the Sulong bitcode library provides a `_start` function which sets up `argc`/`argv`/`env` and calls `main`. This change is necessary to support a bitcode version of a libc in the future.

Additionally, `exit`/`atexit` is implemented in bitcode and unsupported/dangerous functions (e.g. `fork`/`vfork`/`exec`*/`setjmp`/`longjmp`/…) are implemented as bitcode functions which print an error and `abort()`.

This PR allows to run most of the [coreutils](https://www.gnu.org/software/coreutils/coreutils.html) and most [busybox](https://busybox.net/) applets, including `ls`, `du`, `df`, `ip`, `ed`, `vi`, `wget`, `md5sum`, `bzip2` and many more. There are still problems though, because `setjmp`/`longjmp`/`fork`/`vfork`/`exec`* is not supported and it looks like `errno` might confuse the JVM sometimes.

It is also possible to run very basic toy programs like a simple `cat` with a [bitcode version of musl libc](https://github.com/pekd/sulong-musl/tree/sulong).


projects/com.oracle.truffle.llvm.libraries.bitcode/src/crt0.c
```diff
@@ -0,0 +1,40 @@
...
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <stdlib.h>
+
+int main(int argc, char **argv, char **envp);
```
anatol1234: This does not match the signature of the Rust main function, which takes an i64 as argument count. A `ClassCastException` will be thrown on the attempt to pass the integer argc for the long argc to the Rust main function. Maybe choosing `_start` as entry point by default is overall too C specific. Could you generalize your code in a way that it will not break support for other llvm languages like Rust? - for example, using the old implementation as fallback if the signature of the main function does not match.

pekd: Basically `_start` has nothing to do with C at all. It's the entry point of every executable on Linux. The value of `_start` is used for the `e_entry` field in the ELF header. If no `_start` is found, the linker prints a warning and defaults to the platform's default start address. On x86_64 this might look like this:
```
/usr/bin/ld: warning: cannot find entry symbol _start; defaulting to 0000000000000310
```
In C, the `_start` label is [defined in the libc](https://git.musl-libc.org/cgit/musl/tree/arch/x86_64/crt_arch.h). The `_start` code in the libc normally sets up `argc` and `argv` and [then calls some entry point of the libc](https://git.musl-libc.org/cgit/musl/tree/crt/crt1.c), which then [initializes various things (including TLS) and then calls `main`](https://git.musl-libc.org/cgit/musl/tree/src/env/__libc_start_main.c).

To be precise: `_start` is not even a function. It is just a label where execution starts. The single "argument" in this bitcode implementation is normally retrieved by assembler code, because it's the stack pointer when execution starts at `_start`. On Linux the stack pointer points to `argc`, followed by `argv`, followed by `env`, followed by `auxv`, followed by auxv data, followed by argv strings, followed by env strings (followed by the execfn string). Therefore, you cannot `return` from `_start` (because there is no return address) and you have to call the `SYS_exit` syscall to terminate the program.

`main` *is* C specific and defined in the C standard. That's why you can define your own `_start` if you don't use a libc. That's what languages like Rust do in their runtime libraries.

The only question is how to distinguish languages which use a C style `main` from other languages like Rust which don't use a C style `main`.

anatol1234: But once it is compiled to llvm IR, `_start` ends up to be just an ordinary function, right? Anyways, it shouldn't be a problem if your implementation of `_start` became the entry point for every llvm language executed on top of Sulong. The problem is just the signature of the main function. Therefore, I would say, either we have a `_start` that can handle any main function signature specified in the bitcode that gets executed or we use the old approach which calls the main function directly as fallback in the case that Sulong encounters a non C style main.

projects/com.oracle.truffle.llvm.asm.amd64/src/com/oracle/truffle/llvm/asm/amd64/AsmFactory.java
```diff
@@ -1008,6 +1023,168 @@ void createBinaryOperationImplicitSize(String operation, AsmOperand a, AsmOperan
                     throw new AsmParseException("invalid operand type: " + dstType);
                 }
                 break;
+            case "bsr":
```
mrigger: Could we (for the next PR) implement an interface for the creation of inline assembly instructions, similarly as we do in the `NodeFactory`?

projects/com.oracle.truffle.llvm.asm.amd64/src/com/oracle/truffle/llvm/asm/amd64/AsmFactory.java
```diff
@@ -1008,6 +1023,168 @@ void createBinaryOperationImplicitSize(String operation, AsmOperand a, AsmOperan
...
+                    }
+                    statements.add(out);
+                    return;
+                } else {
+                    throw new AsmParseException("invalid operand type: " + dstType);
+                }
+            }
+            case "cmpxchg": {
+                if (dstType instanceof PointerType) {
+                    // dstType = PrimitiveType.I64;
```
mrigger: Is this dead code or a comment?

projects/com.oracle.truffle.llvm.asm.amd64/src/com/oracle/truffle/llvm/asm/amd64/AsmFactory.java
```diff
@@ -1511,6 +1688,7 @@ private void getArguments() {
             this.result = new StructLiteralNode(retOffsets, writeNodes, valueNodes, addr);
         }
 
+        todoRegisters.remove("rsp");
```
mrigger: Why do we remove `rsp`?

projects/com.oracle.truffle.llvm.libraries.bitcode/src/crt0.c
```diff
@@ -0,0 +1,55 @@
...
+  int argc = p[0];
+  char **argv = (void *)(p + 1);
+  char **envp = argv + argc + 1;
+
+  switch (type) {
+  /* C/C++/... */
+  default:
+  case 0:
+    exit(main(argc, argv, envp));
+    return 0;
```
mrigger: Do you need `return 0` here? `exit` should be declared as `noreturn` (see https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html), so the compiler should know that `return 0` is never reached. I think it would be clearer to remove it.

projects/com.oracle.truffle.llvm.libraries.bitcode/src/exit.c
```diff
@@ -0,0 +1,73 @@
...
+  void *arg;
+};
+
+static struct entry head = { NULL, NULL, NULL };
+
+static void __funcs_on_exit() {
+  struct entry *entry = &head;
+  while (entry->next) {
+    entry = entry->next;
+    entry->func(entry->arg);
```
mrigger: Can we free each entry on exit? This would help memory leak detectors.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/support/LLVMAMD64String.java
```diff
@@ -0,0 +1,112 @@
...
+import com.oracle.truffle.llvm.runtime.memory.LLVMMemory;
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMExpressionNode;
+
+public abstract class LLVMAMD64String extends LLVMExpressionNode {
+    private static Constructor<? extends ByteBuffer> directByteBufferCtor;
+
+    static {
+        try {
+            @SuppressWarnings("unchecked")
+            Class<? extends ByteBuffer> clazz = (Class<? extends ByteBuffer>) Class.forName("java.nio.DirectByteBuffer");
```
mrigger: Can you document why you are using reflection to access the byte buffer?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64File.java
```diff
@@ -0,0 +1,337 @@
...
+    private static FileChannel getInputChannel(int fd) {
+        return new FileInputStream(getFd(fd)).getChannel();
+    }
+
+    private static FileChannel getOutputChannel(int fd) {
+        return new FileOutputStream(getFd(fd)).getChannel();
+    }
+
+    @TruffleBoundary
+    public static int open(String filename, int flags, @SuppressWarnings("unused") int mode) {
```
mrigger: I think it would be better if we put each method in a separate node. Then we could use profiling, for example, here for `rdwr`. Additionally, we could use `TruffleBoundary` on a more fine-grained basis to prevent, for example, that `LLVMAddress` has to be allocated on the heap. This applies not only to this class, but also to most of the other operation which could be implemented as nodes.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Memory.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static final int MAP_SHARED = 0x01;
+    public static final int MAP_PRIVATE = 0x02;
+    public static final int MAP_TYPE = 0x0f;
+    public static final int MAP_FIXED = 0x10;
+    public static final int MAP_ANONYMOUS = 0x20;
+    public static final int MAP_UNINITIALIZED = 0x4000000;
+
+    public static long brk(@SuppressWarnings("unused") LLVMAddress ptr) {
+        return -LLVMAMD64Error.ENOSYS;
```
mrigger: Could you document the current assumptions if a system does not implement the expected semantics (in this and other cases)?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Path.java
```diff
@@ -0,0 +1,41 @@
...
+package com.oracle.truffle.llvm.nodes.asm.syscall;
+
+import java.nio.file.Paths;
+
+import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
+
+public class LLVMAMD64Path {
+    @TruffleBoundary
+    public static String getcwd() {
+        return Paths.get(".").toAbsolutePath().normalize().toString();
```
mrigger: Is this used at several places? If not, you could inline the method.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64ProcessStat.java
```diff
@@ -0,0 +1,93 @@
...
+    private final long pid;
+    private final String comm;
+    private final char state;
+    private final long ppid;
+    private final long pgrp;
+    private final long session;
+    private final long ttyNr;
+    private final long tpgid;
+
+    public LLVMAMD64ProcessStat(String stat) {
```
mrigger: What is `stat`?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallNode.java
```diff
@@ -0,0 +1,194 @@
...
+                throw new RuntimeException("unknown syscall " + rax);
+        }
+    }
+
+    @SuppressWarnings("unused")
+    @Specialization
+    protected long executeI64(long rax, long rdi, LLVMAddress rsi, long rdx, long r10, long r8, long r9) {
+        switch ((int) profile.profile(rax)) {
+            case LLVMAMD64Syscall.SYS_read: {
+                // byte[] buf = new byte[(int) rdx];
```
mrigger: Can we remove the dead code or document why it is there (also below)?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallNode.java
```diff
@@ -0,0 +1,194 @@
...
+
+@NodeChildren({@NodeChild("rax"), @NodeChild("rdi"), @NodeChild("rsi"), @NodeChild("rdx"), @NodeChild("r10"), @NodeChild("r8"), @NodeChild("r9")})
+public abstract class LLVMAMD64SyscallNode extends LLVMExpressionNode {
+    private final LongValueProfile profile = LongValueProfile.createIdentityProfile();
+
+    protected static void exit(int code) {
+        throw new LLVMExitException(code);
+    }
+
+    @Specialization
```
mrigger: We want to have nodes for each operation (see above). You can use a `@Cached` annotation for a lookup of the node and just always pass all arguments. What do you think @grimmerm?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/support/LLVMAMD64String.java
grimmerm: Use of reflection should be avoided at all costs

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Info.java
```diff
@@ -0,0 +1,136 @@
...
+        machine = arch;
+    }
+
+    private static String readFile(String name, String fallback) {
+        try {
+            Path path = Paths.get(name);
+            if (Files.exists(path)) {
+                return new String(Files.readAllBytes(path)).trim();
+            }
+        } catch (Exception e) {
```
grimmerm: also add transferToInterpreter to empty catch clauses

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallNode.java
grimmerm: agreed

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Memory.java
pekd: Correct code has to check for errors after every syscall. If a syscall is not implemented, it has to return ENOSYS. Therefore, this is the correct behavior for unimplemented syscalls. Real world applications ignore that sometimes, and things will break.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64ProcessStat.java
pekd: The contents of `/proc/self/stat`.

mikehearn:
Does this obsolete my PR #753 ? Atexit has Java implementations as well but this PR changes a C implementation.

grimmerm:
@mikehearn probably yes.
Once that is merged, can you pls verify if that fixes your issue? Thx

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/LLVMAMD64AdcNode.java
```diff
@@ -53,8 +93,21 @@ public LLVMAMD64AdcbNode(LLVMAMD64UpdateCPZSOFlagsNode flags) {
         protected byte executeI8(VirtualFrame frame, byte left, byte right, boolean cf) {
             byte c = (byte) (cf ? 1 : 0);
             byte result = (byte) (left + right + c);
-            boolean overflow = (result < 0 && left > 0 && right > 0) || (result > 0 && left < 0 && right < 0);
-            boolean carry = ((left < 0 || right < 0) && result > 0) || (left < 0 && right < 0);
+            boolean overflow;
+            boolean carry;
+            if (!cf) {
```
mrigger: We should profile these conditions (here and in the other execute methods of this class).

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Info.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @TruffleBoundary
+    public static String getDomainName() {
+        return readFile("/proc/sys/kernel/domainname", null);
+    }
+
+    public static long uname(LLVMAddress name) {
+        LLVMAddress ptr = name;
+        LLVMAMD64String.strcpy(ptr, sysname);
+        ptr = ptr.increment(65);
```
mrigger: Why `65`?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Memory.java
mrigger: Could you add some comment like "// not supported"?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Memory.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @SuppressWarnings("unused")
+    public static long mmap(LLVMAddress addr, long len, int prot, int flags, int fildes, long off) {
+        if ((flags & MAP_ANONYMOUS) != 0) {
```
mrigger: We can profile the flag here.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Info.java
pekd: Because the member is 65 bytes long, see notes section of the [uname function](http://man7.org/linux/man-pages/man2/uname.2.html).

mrigger: Can you quote the relevant paragraph, add a link, and introduce a constant for that?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMd64Time.java
```diff
@@ -0,0 +1,87 @@
...
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.nodes.asm.syscall;
+
+import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
+import com.oracle.truffle.llvm.runtime.LLVMAddress;
+import com.oracle.truffle.llvm.runtime.memory.LLVMMemory;
+
+public class LLVMAMd64Time {
```
rschatz: Typo (`AMd64Time`, should be `AMD64Time`)

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallIoctlNode.java
```diff
@@ -0,0 +1,50 @@
...
+
+public abstract class LLVMAMD64SyscallIoctlNode extends LLVMAMD64SyscallOperationNode {
+    public LLVMAMD64SyscallIoctlNode() {
+        super("ioctl");
+    }
+
+    @Specialization
+    protected long executeI64(long fd, long request, LLVMAddress argp) {
+        return LLVMAMD64File.ioctl((int) fd, (int) request, argp.getVal()); // TODO: implement
+                                                                            // properly
```
grimmerm: what is missing here?
Can you add more details about what "a proper implementation" should look like

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallNode.java
```diff
@@ -0,0 +1,160 @@
...
+    private boolean traceEnabled() {
+        if (traceStream == null) {
+            CompilerDirectives.transferToInterpreterAndInvalidate();
+            traceStream = SulongEngineOption.getStream(getContext().getEnv().getOptions().get(SulongEngineOption.DEBUG));
+            traceEnabledFlag = SulongEngineOption.isTrue(getContext().getEnv().getOptions().get(SulongEngineOption.DEBUG));
+        }
+        return traceEnabledFlag;
+    }
+
+    private PrintStream traceStream() {
```
grimmerm: duplicated code

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallOperationNode.java
```diff
@@ -0,0 +1,46 @@
...
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.nodes.asm.syscall;
+
+import com.oracle.truffle.api.nodes.Node;
+
+public abstract class LLVMAMD64SyscallOperationNode extends Node {
+    private String name;
```
grimmerm: why is this field not final?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallIoctlNode.java
pekd: According to the man page:
> The type of `arg` depends upon the particular control request, but it shall be either an integer or a pointer to a device-specific data structure.

It's not possible to decide here if the last argument is a pointer or an integer, but it doesn't matter as long as the ioctl commands are not implemented.

grimmerm: thanks for the clarification

pekd:
I'll close this PR now because most changes are already in master.

PR #775, opened 2017-09-18
Add Mach-O support to Sulong.

rmosaner:
This pull requests adds support for Mach-O files to Sulong. Using clang's "-fembed-bitcode" flag, the bitcode can be injected into the compiled executable.

However, most executables are built from multiple source files where compiling with "-fembed-bitcode" results in all bitcodes being stored in a Xar archive without being linked. To avoid a runtime dependency to llvm-link, a bitcode injector tool has been added which requires an executable being built with WLLVM and then automatically adds the linked bitcode into the WLLVM Mach-O executable.
To call the bitcode injector, one can use mx:

`mx inject-bitcode wllvm.out`  ... injects the bitcode in-place
`mx inject-bitcode wllvm.out injected.out` ... injects the bitcode into an external file


projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/machO/MachOFile.java
```diff
@@ -0,0 +1,289 @@
...
+            FileChannel out = new FileOutputStream(file).getChannel();
+            files.add(file);
+            out.write(bitcodes[i]);
+            out.close();
+        }
+
+        // create llvm-link command
+        File bcFile = Files.createTempFile(".linkedBc_", ".bc").toFile();
+        bcFile.deleteOnExit();
+        StringBuilder cmd = new StringBuilder("llvm-link -o ");
```
jkreindl: I don't think its a good idea to call `llvm-link` directly from java. This doesn't check version compatibility like `mx` does which could lead to problems when homebrew updates to a newer llvm version. As a short-term solution we could just throw an exception in this case, telling users to recompile with WLLVM. Also, since we already support libraries, I think it would be possible to just pass each of the `ByteBuffer` to the parser sequentially. We may need to implement logic to find / keep the program entry point in this case, but I think this would be preferable over having this dependency.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/machO/MachODylibCommand.java
```diff
@@ -0,0 +1,93 @@
...
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.parser.machO;
+
+import java.nio.ByteBuffer;
+
+public class MachODylibCommand extends MachOLoadCommand {
```
jkreindl: This and `MachOSection` could be declared final.

tests/com.oracle.truffle.llvm.tests.wllvm/Makefile
```diff
@@ -44,8 +44,10 @@ else
...
 	$(QUIETLY) extract-bc --output $@ $<
 
-$(OBJDIR)/ref.out: $(OBJDIR)/ref.wllvm $(OBJDIR)/wllvm.bc
-	cp $< $@
-	$(QUIETLY) objcopy -I binary -O elf64-x86-64 -B i386:x86-64 --rename-section .data=.llvmbc $(OBJDIR)/wllvm.bc $@	
+$(OBJDIR)/wllvm.bc.o: $(OBJDIR)/wllvm.bc
+	$(QUIETLY) objcopy -I binary -O elf64-x86-64 -B i386:x86-64 --rename-section .data=.llvmbc $< $@
+
+$(OBJDIR)/ref.out: caller.cpp callee.cpp $(OBJDIR)/wllvm.bc.o
+	clang++ -o $@ $^
```
jkreindl: Can you prepend $(QUIETLY) to this? It keeps `make` from printing the command to stdout unless running `mx -v`.

mx.sulong/mx_sulong.py
```diff
@@ -363,6 +363,11 @@ def runLLVM(args=None, out=None):
     vmArgs, sulongArgs = truffle_extract_VM_args(args)
     return mx.run_java(getCommonOptions(False) + vmArgs + getClasspathOptions() + ["com.oracle.truffle.llvm.Sulong"] + sulongArgs, out=out)
 
+def injectBitcode(args=None, out=None):
+    """Injects the bitcode of a Mach-O WLLVM build into the executable"""
+    vmArgs, injectorArgs = truffle_extract_VM_args(args)
+    return mx.run_java(getCommonOptions(False) + vmArgs + mx.get_runtime_jvm_args('BITCODE_INJECTOR') + ["com.oracle.truffle.llvm.parser.macho.injector.Injector"] + injectorArgs, out=out)
```
mrigger: Do we actually need the `getCommonOptions` here, since the code is not really Truffle/Graal related?

mx.sulong/mx_sulong.py
```diff
@@ -675,4 +680,5 @@ def getResults(self):
     'lli' : [runLLVM, ''],
     'travis1' : [travis1, ''],
     'travis2' : [travis2, ''],
+    'inject' : [injectBitcode, ''],
```
mrigger: Maybe call the command `inject-bc` or `inject-bitcode` so that its purpose is self-explanatory?

projects/com.oracle.truffle.llvm.parser.macho.injector/src/com/oracle/truffle/llvm/parser/macho/injector/Injector.java
```diff
@@ -0,0 +1,173 @@
...
+        File bcFile = Files.createTempFile("extractedBc_", ".bc").toFile();
+        bcFile.deleteOnExit();
+
+        extractBc(wllvmFile, bcFile);
+
+        // read wllvm built Mach-O executable
+        ByteBuffer machOBuffer = readFile(wllvmFile);
+        MachOFile machO = MachOFile.create(machOBuffer);
+
+        MachOSegmentCommand seg = machO.getSegment("__WLLVM");
```
mrigger: Having one (or more) helper methods would increase the readability. For example, you could have a method `getBitcodeSection` that gets the segment and section and performs checks on them.

projects/com.oracle.truffle.llvm.parser.macho.injector/src/com/oracle/truffle/llvm/parser/macho/injector/Injector.java
```diff
@@ -0,0 +1,173 @@
...
+
+        if (sec == null) {
+            throw new RuntimeException("Not a valid wllvm build (missing __llvm_bc section)!");
+        }
+
+        // zero out the paths of the bc files
+        zeroOutSection(machOBuffer, sec);
+
+        int secOffset = sec.getCmdOffset();
+        int offsetOffset = secOffset + MachOSegmentCommand.SEGNAME_SIZE + MachOSegmentCommand.SECTNAME_SIZE + Long.SIZE / Byte.SIZE + Long.SIZE / Byte.SIZE;
```
mrigger: You can use `Long.BYTES` instead of `Long.SIZE / Byte.SIZE`.

mrigger (commented):
In general, the PR looks good to me! I added a few suggestions for improvements.

zapster:
I've picked up most of the changes from this PR as the basis for my Mach-O support feature, which has now be [merged into the oracle/graal master branch](https://github.com/oracle/graal/commit/a237181bb5efe0f6bc1eeeec42e3b4b012abe8c9). Thanks for the contribution!

PR #778, opened 2017-10-11
Fix endianess bug in 80bit float

pekd:
… and add `fmodl` intrinsic.

### DIFF
--- a/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/floating/LLVM80BitFloat.java
+++ b/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/floating/LLVM80BitFloat.java
@@ -439,6 +439,17 @@ public final class LLVM80BitFloat {
         return Arrays.hashCode(getBytes());
     }
 
+    public byte[] getBytesBigEndian() {
+        ByteBuffer bb = ByteBuffer.allocate(BYTE_WIDTH);
+        bb.order(ByteOrder.BIG_ENDIAN);
+        short signWithExponent = getExponent();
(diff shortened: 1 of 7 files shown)


projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/floating/LLVM80BitFloat.java
```diff
@@ -448,16 +450,39 @@ public int hashCode() {
         return bb.array();
     }
 
-    public static LLVM80BitFloat fromBytes(byte[] bytes) {
+    public byte[] getBytes() {
+        ByteBuffer bb = ByteBuffer.allocate(BYTE_WIDTH);
+        bb.order(ByteOrder.LITTLE_ENDIAN);
```
rschatz: Do we really need `LITTLE_ENDIAN` here? If we don't care, I think it's better to have the platform endianness, since it's faster.

pekd: If this code is run on a big endian machine, the following `putShort`/`putLong` have to be swapped. Since the whole data type is called `X86_FP80` it's an x86 specific data type that should never exist in non-x86 bitcode (therefore it should never exist in bitcode that runs on a big endian machine).

rschatz: Good point. We might at some point want to be able to run foreign bitcode, but not yet, and anyway in that scenario we will have tons of endian conversions all over the place anyway, so this one is probably not going to hurt much...

rschatz:
Lgtm!

PR #787, opened 2017-10-18
Linux/x86_64 syscalls

pekd:


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64SyscallNode.java
```diff
@@ -0,0 +1,164 @@
...
+                    @Cached("createNode(rax)") LLVMAMD64SyscallOperationNode node, @SuppressWarnings("unused") @Cached("rax") long cachedRax) {
+        if (traceEnabled()) {
+            trace(node);
+        }
+        return node.execute(rdi, rsi, rdx, r10, r8, r9);
+    }
+
+    @SuppressWarnings("unused")
+    @Specialization(replaces = "cachedSyscall")
+    protected long executeI64(long rax, Object rdi, Object rsi, Object rdx, Object r10, Object r8, Object r9) {
```
grimmerm: for now, you can just jump back to the interpreter, create a node on demand and execute it

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Time.java
```diff
@@ -0,0 +1,87 @@
...
+    private static long currentTimeMillis() {
+        return System.currentTimeMillis();
+    }
+
+    @TruffleBoundary
+    private static long nanoTime() {
+        return System.nanoTime();
+    }
+
+    public static int clockGetTime(int clkId, LLVMAddress timespec) {
```
rschatz: I think functions like this should be actually implemented either as a node, or even in C. The advantage would be that `timespec` could then also be a foreign object that looks like a `struct timespec`, instead of a raw memory pointer.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/posix/LLVMAMD64PosixCallNode.java
```diff
@@ -0,0 +1,77 @@
...
+
+    // Workaround for nice syntax + Truffle DSL
+    public final Object execute(Object... args) {
+        return executeObject(args);
+    }
+
+    public abstract Object executeObject(Object[] args);
+
+    @Specialization
+    public Object execute(Object[] args, @Cached("createFunction()") TruffleObject function) {
```
rschatz: Having two functions with the same name, one of them `Object...`, is dangerous.
Perhaps just rename this to `doCall`.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/posix/LLVMAMD64PosixCallNode.java
```diff
@@ -0,0 +1,77 @@
...
+    @Child private Node nativeExecute;
+
+    public LLVMAMD64PosixCallNode(String name, String signature, int args) {
+        this.name = name;
+        this.signature = signature;
+        nativeExecute = Message.createExecute(args).createNode();
+    }
+
+    protected TruffleObject createFunction() {
+        return getContext().getNativeLookup().getNativeFunction("@__sulong_posix_" + name, String.format(signature));
```
rschatz: Why `String.format`?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/asm/syscall/LLVMAMD64Error.java
```diff
@@ -0,0 +1,176 @@
...
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.nodes.asm.syscall;
+
+public class LLVMAMD64Error {
```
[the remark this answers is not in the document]
pekd: POSIX only specifies symbolic names for error values. Linux choses numbers for these symbolic names. There are even differences between Linux different architectues:
```c
#include <errno.h>
#include <stdio.h>

int main(void)
{
	printf("EDEADLOCK = %d\n", EDEADLOCK);
	return 0;
}
```
On Linux/x86_64, this outputs 35, on Linux/ppc this outputs 58.

For `LLVMAMD64String` and `LLVMAMD64File`: should they also be moved to a different package?

projects/com.oracle.truffle.llvm.libraries.bitcode/src/syscall.c
```diff
@@ -0,0 +1,47 @@
...
+__attribute__((weak)) int64_t syscall(int64_t n, ...) {
+  va_list ap;
+  int64_t a, b, c, d, e, f;
+  va_start(ap, n);
+  a = va_arg(ap, int64_t);
+  b = va_arg(ap, int64_t);
+  c = va_arg(ap, int64_t);
+  d = va_arg(ap, int64_t);
+  e = va_arg(ap, int64_t);
+  f = va_arg(ap, int64_t);
```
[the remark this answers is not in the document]
pekd: AFAIK Sulong allocates a memory area for the GPRs, but only writes values there if they are available. `va_arg` reads from that area and may read garbage if nothing was written there. I will implement an intrinsic for the `syscall` function to avoid another problem: arguments are read as `int64_t` but might have been passed as something different (e.g. `short`). In this case, the value has to be sign extended to 32bit (that's the de-facto ABI on x86_64) and when a 32bit register is written, it is zero extended to 64bit. Since the call site doesn't know the prototype of the syscall, something like `syscall(something, -1)` might produce wrong results right now.

PR #788, opened 2017-10-18
Entry point

pekd:
Set entry point to `_start`

### DIFF
--- a/projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/LLVMRootNodeFactory.java
+++ b/projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/LLVMRootNodeFactory.java
@@ -29,40 +29,24 @@
  */
 package com.oracle.truffle.llvm.parser.factories;
 
-import java.util.Arrays;
-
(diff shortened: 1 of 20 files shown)


projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/LLVMRootNodeFactory.java
```diff
@@ -137,4 +107,88 @@ private static LLVMAddress getArgsAsStringArray(Object[] args) {
...
+        LLVMAddress ptr = address;
+        for (byte b : string.getBytes()) { // automatic charset conversion
+            LLVMMemory.putI8(ptr, b);
+            ptr = ptr.increment(1);
+        }
+        LLVMMemory.putI8(ptr, (byte) 0);
+        return ptr.increment(1);
+    }
+
+    private static LLVMAddress getArgs(Object[] mainArgs) {
```
grimmerm: This function has to be modeled with Sulong nodes.
Use a malloc node to allocate memory, use the free node to free it.
Use the pointer arithmetic nodes to compute address values. 
Use memory access nodes to load/store data to memory.

PR #799, opened 2017-10-31
Add support for TLS (x86_64 inline asm)

pekd:
Allow reads from `fs:0` and add syscall to set `fs:0`.


projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/LLVMContext.java
```diff
@@ -91,6 +91,9 @@
 
     private final List<ContextExtension> contextExtension;
 
+    private final ThreadLocal<LLVMAddress> tls = ThreadLocal.withInitial(LLVMAddress::nullPointer);
```
mrigger: @grimmerm Should this be a `ContextExtension`?

Issue #801, opened 2017-11-02
Cannot return types other than i32 from main


pointhi:
The following snippet:

```c
short main() {
    return 0 ;
}
```

compiles correctly (while outputing a warning because of the return type) to the following ir:

```llvm
define signext i16 @main() {
  ret i16 0
}
```

There seems to be a regression in sulong since a few days, because this snippet is no longer able to be executed:

```
C stack trace:
	 LLVM IR Function @_start in libsulong.bc in Block {id: 4 name: implicit21}

	at com.oracle.truffle.llvm.nodes.base.LLVMBasicBlockNode.executeStatements(LLVMBasicBlockNode.java:114)
	at com.oracle.truffle.llvm.nodes.control.LLVMDispatchBasicBlockNode.executeGeneric(LLVMDispatchBasicBlockNode.java:75)
	at com.oracle.truffle.llvm.nodes.func.LLVMFunctionStartNode.execute(LLVMFunctionStartNode.java:74)
	at com.oracle.truffle.api.impl.DefaultCallTarget.callDirectOrIndirect(DefaultCallTarget.java:64)
	at com.oracle.truffle.api.impl.DefaultDirectCallNode.call(DefaultDirectCallNode.java:43)
	at com.oracle.truffle.llvm.nodes.func.LLVMGlobalRootNode.executeIteration(LLVMGlobalRootNode.java:131)
	at com.oracle.truffle.llvm.nodes.func.LLVMGlobalRootNode.execute(LLVMGlobalRootNode.java:94)
	at com.oracle.truffle.api.impl.DefaultCallTarget.callDirectOrIndirect(DefaultCallTarget.java:64)
	at com.oracle.truffle.api.impl.DefaultDirectCallNode.call(DefaultDirectCallNode.java:43)
	at com.oracle.truffle.llvm.nodes.base.LLVMMainFunctionReturnValueRootNode$LLVMMainFunctionReturnNumberRootNode.execute(LLVMMainFunctionReturnValueRootNode.java:84)
	at com.oracle.truffle.api.impl.DefaultCallTarget.call(DefaultCallTarget.java:81)
	at com.oracle.truffle.api.vm.PolyglotContextImpl.eval(PolyglotContextImpl.java:645)
	at org.graalvm.polyglot.Context.eval(Context.java:148)
	at com.oracle.truffle.llvm.Sulong.executeMain(Sulong.java:142)
	at com.oracle.truffle.llvm.Sulong.main(Sulong.java:134)
[...]

PR #805, opened 2017-11-03
Support non-standard main return types

pekd:
Quick fix for #801

In the future, the parser has to automatically add bitcasts if types don't match.


mrigger:
LGTM!

PR #810, opened 2017-11-15
Disable pthreads

pekd:
Add stubs for `clone` and pthreads functions to prevent unsupported and dangerous operations.


mrigger:
LGTM

PR #812, opened 2017-11-15
Remove pulldragonegg from travis config

pekd:


rschatz:
Thanks for fixing that.
Regarding the failing `gcc_fortran` tests: I'm going to prepare a pre-built dragonegg download so we can run these tests on travis. Can you please disable this gate run for now?

PR #813, opened 2017-11-17
Some simple Metadata Parser improvements

pointhi:
* mark some constants as final
* combine multiple separated DwarfTag Enums into one big one

### DIFF
--- a/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/metadata/MDType.java
+++ b/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/metadata/MDType.java
@@ -29,7 +29,11 @@
  */
 package com.oracle.truffle.llvm.parser.metadata;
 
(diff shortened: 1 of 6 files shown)


projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/metadata/MDType.java
```diff
@@ -83,4 +92,47 @@ public void replace(MDBaseNode oldValue, MDBaseNode newValue) {
             file = newValue;
         }
     }
+
+    // @see: https://github.com/llvm-mirror/llvm/blob/master/include/llvm/BinaryFormat/Dwarf.def
+    public enum DwarfTag {
+        UNKNOWN(-1),
+
+        DW_TAG_ARRAY_TYPE(0x1),
```
jkreindl: Just for clarity, could you add comments for which tags correspond to which type class (basic type, derived type or composite type)?

PR #817, opened 2017-11-26
Add support for Rust test cases

anatol1234:


mx.sulong/suite.py
```diff
@@ -388,6 +397,7 @@
         "com.oracle.truffle.llvm.tests.nfi",
         "com.oracle.truffle.llvm.tests.sulong",
         "com.oracle.truffle.llvm.tests.sulongcpp",
+	"com.oracle.truffle.llvm.tests.sulongrust",
```
pekd: Don't mix spaces and tabs.

.travis.yml
```diff
@@ -79,7 +82,7 @@ matrix:
     jdk: oraclejdk8
   - env: GATE_TAGS='gcc_cpp' COMPILE_COMMAND='mx build --dependencies SULONG_TEST'
     jdk: oraclejdk8
-  - env: DOWNSTREAM_REPO='https://github.com/graalvm/graal.git' DOWNSTREAM_SUITE='substratevm' DOWNSTREAM_COMMAND='--dynamicimport sulong gate --tags build,sulong'
+  - env: DOWNSTREAM_REPO='https://github.com/graalvm/graal.git' DOWNSTREAM_SUITE='substratevm' DOWNSTREAM_COMMAND='--dynamicimport sulong gate --tags build,sulong' SULONG_USE_RUSTC='false'
```
rschatz: Why does it not work with substratevm? I think we should rather fix the bug than disable the test.
If it doesn't work with substratevm, it also won't work in the default configuration of the graalvm distribution.

anatol1234: The problem is that somewhere the Rust standard library needs to be linked. I configure this in mx_sulong. Does graal dodge the `_unittest_config_participant` hook in the default configuration?

rschatz: That's probably a problem with the test harness then. I guess the arguments are not being forwarded correctly between the VM running the test harness and the Sulong VM. I'll take a look at that.

anatol1234: Forwarding of arguments has been functioning now for a while. I configured the arguments in `testLLVMImage` and reactivated Rust for the graalvm gate (formerly substratevm gate). 

The changes revolving around `LLVMGlobal` broke support for the recent Rust versions depending on `@std::rt::lang_start_internal`. I fixed the intrinsic accordingly and  added a gate check for Rust 1.27.1.

PR #818, opened 2017-11-29
Add partial support for function vectors

pekd:


projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/vector/LLVMFunctionVector.java
```diff
@@ -0,0 +1,128 @@
...
+    public Object getValue(int index) {
+        return vector[index];
+    }
+
+    public LLVMAddress getAddress(int index) {
+        if (vector[index] instanceof LLVMAddress) {
+            return (LLVMAddress) vector[index];
+        } else if (vector[index] instanceof LLVMFunctionDescriptor) {
+            LLVMFunctionDescriptor fd = (LLVMFunctionDescriptor) vector[index];
+            LLVMAddress ptr = LLVMAddress.fromLong(fd.toNative().asPointer());
```
rschatz: Please don't introduce new users of `toNative()`, especially on the fast-path. We're currently trying to get rid of usages of this function. See https://github.com/graalvm/sulong/blob/master/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/LLVMFunctionDescriptor.java#L479

The proper way to do this on the fast-path would be to use a `LLVMToNativeNode`. Then you also wouldn't need the `if` cascade, the `LLVMToNativeNode` would automatically specialize itself to the correct type.

projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/memory/LLVMMemory.java
```diff
@@ -490,6 +501,14 @@ public void putVector(LLVMAddress address, LLVMAddressVector vector) {
         }
     }
 
+    public void putVector(LLVMAddress address, LLVMFunctionVector vector) {
+        long currentPtr = address.getVal();
+        for (int i = 0; i < vector.getLength(); i++) {
+            putAddress(currentPtr, vector.getAddress(i));
```
rschatz: See other comment in `getAddress`. Instead, use a `LLVMToNativeNode`. You can pass the node in as argument to the function, and create it using `@Cached("createToNativeWithTarget()")` at the callsite.

PR #834, opened 2018-02-12
Fix logo designer website link

ot:
There's a missing dash in the URL, which directs to the wrong user profile.


rschatz:
Thanks for pointing this out. Should be fixed now.

PR #843, opened 2018-05-14
Add support for Rust version 1.26.0

anatol1234:
The signature of the `std::rt::lang_start` function has changed since version 1.19.0 to support returning a `Result` from the main function. This feature has been stabilized in version 1.26.0. This pull request includes support for the new signature. It enables Sulong to run code like:

```Rust
fn main() -> Result<(), &'static str> {
    Err("something went wrong")
}
```
Output: `Error: "something went wrong"`


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/rust/LLVMStart.java
```diff
@@ -0,0 +1,202 @@
...
+    public abstract static class LLVMLangStartInternal extends LLVMStart {
+
+        protected LangStartVtableType createLangStartVtable(Type vtableType) {
+            DataLayout dataSpecConverter = getContextReference().get().getDataSpecConverter();
+            return LangStartVtableType.create(dataSpecConverter, vtableType);
+        }
+
+        @Specialization
+        @SuppressWarnings("unused")
+        protected long doOp(StackPointer stackPointer, LLVMNativePointer mainPointer, LLVMGlobal vtable, long argc, LLVMPointer argv,
```
christianhaeubl: This code seems to rely on the fact that it is executed only once (you are caching values that are derived from the method arguments, without using any guards).
- if it is guaranteed that this code is only executed once, the caching will only be an overhead and won't have any benefit as the code is never compiled
- if it is not guaranteed that this code is only executed once, you will need guards

christianhaeubl:
Thanks for the contribution, I added one comment regarding LLVMLangStartInternal.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/intrinsics/rust/LLVMStart.java
anatol1234: It indeed relies on the fact that it is executed only once. `lang_start_internal` is called exactly once at program start. I removed most of the caching in my latest commit. I am unsure whether it makes sense to cache `LLVMMemory`. Its cached value is not derived from method arguments, but I guess I should still get rid of it for the reason that it is mere overhead.

christianhaeubl: Thanks for cleaning that up. Yes, you could also get rid of caching `LLVMMemory`. Besides that, the change looks good to me.

Issue #840, opened 2018-04-18
Incorrect allocation of struct literals and returned structs


anatol1234:
I have reason to suspect that structs which are returned by value and also struct literals are allocated incorrectly by Sulong with respect to the LLVM language specification.

Currently, Sulong inserts a struct allocating `alloca` instruction for functions and instructions which return structs by value. Therefore, LLVM stack memory occupied with returned structs is not freed until the containing function returns. This approach apparently fails if structs are returned within loops: the stack memory grows with every loop iteration and in the worst case causes a segfault due to stack overflow. I assume Clang emits IR which is less prone to this issue. The Rust compiler, however, uses lots of insertvalue, arithmetic with overflow etc. instructions and builtins on O0.

Here is a trivial safe Rust program which yields a guaranteed segfault after about a minute (rustc 1.19.0, Ubuntu 64-bit):

```Rust
fn main() {
    let c = &mut [0]; 
    loop{
	c[..][0] = 42;
    }
}
```

Compile and run with:
```
rustc --emit=llvm-bc <source.rs>
mx lli <source.bc>
```
Compiled with rustc `--emit=llvm-bc -C inline-threshold=0 <source.rs>` it segfaults almost instantly. 
This is (as I see it) the critical part in the IR (`inline-threshold=0`):

```LLVM
bb1:                                              ; preds = %bb3, %start
  %1 = bitcast [1 x i32]* %_2 to i32*
  %2 = insertvalue { i32*, i64 } undef, i32* %1, 0
  %3 = insertvalue { i32*, i64 } %2, i64 1, 1
  %4 = insertvalue { i32*, i64 } undef, i32* %1, 0
  %5 = insertvalue { i32*, i64 } %4, i64 1, 1
  %6 = extractvalue { i32*, i64 } %5, 0
  %7 = extractvalue { i32*, i64 } %5, 1
  br label %bb2

bb2:                                              ; preds = %bb1
  %8 = icmp ult i64 0, %7
  %9 = call i1 @llvm.expect.i1(i1 %8, i1 true)
  br i1 %9, label %bb3, label %panic

bb3:                                              ; preds = %bb2
  %10 = getelementptr inbounds i32, i32* %6, i64 0
  store i32 42, i32* %10
  br label %bb1
```
`%2, %3, %4` and `%5` are allocated over and over again with every loop iteration and cause a segfault as a consequence.

A minimal (self written) `LLVM` program which reproduces this issue looks like this:
```LLVM
[...]

PR #844, opened 2018-05-30
Fix data stack overflows 

anatol1234:
Fixes issue #840. Programs like the following should not segfault anymore: 

```Rust
fn main() {
    let c = &mut [0]; 
    loop{
	c[..][0] = 42;
    }
}
```

Compile and run with:
```
rustc --emit=llvm-bc <source.rs>
mx lli <source.bc>
```
Each stack frame on a data stack is associated with a region (`UniqueRegion`) reserved for allocated values which are supposed to be unique per frame. The size of each region is calculated ahead of time by evaluating the maximum size uniquely allocated values will take up in the corresponding frame. Also, unique slots holding relative addresses to the frame's unique region base pointer are assigned to unique alloc nodes (`LLVMUniqueAllocConstInstruction`). This way, each region can be allocated at the top of the corresponding frame underneath the parameters before the basic blocks are dispatched.  Unique alloc nodes can then access their stack slots through their assigned relative addresses.

### DIFF
--- a/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/LLVMAllocInstruction.java
+++ b/projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/LLVMAllocInstruction.java
@@ -63,7 +63,7 @@ public abstract class LLVMAllocInstruction extends LLVMExpressionNode {
         return stackPointer;
     }
 
-    public abstract static class LLVMAllocaConstInstruction extends LLVMAllocInstruction {
+    public abstract static class LLVMAllocConstInstruction extends LLVMAllocInstruction {
 
         @CompilationFinal(dimensions = 1) private Type[] types = null;
         @CompilationFinal(dimensions = 1) private int[] offsets = null;
@@ -88,10 +88,23 @@ public abstract class LLVMAllocInstruction extends LLVMExpressionNode {
             return offsets.length;
         }
 
+    }
+
+    public abstract static class LLVMAllocaConstInstruction extends LLVMAllocConstInstruction {
+
         @Specialization
         protected LLVMNativePointer doOp(VirtualFrame frame,
                         @Cached("getLLVMMemory()") LLVMMemory memory) {
             return LLVMNativePointer.create(LLVMStack.allocateStackMemory(frame, memory, getStackPointerSlot(), getSize(), getAlignment()));
+        }
+    }
+
+    public abstract static class LLVMUniqueAllocConstInstruction extends LLVMAllocConstInstruction {
+
+        @Specialization
+        protected LLVMNativePointer doOp(VirtualFrame frame,
+                        @Cached("getLLVMMemory()") LLVMMemory memory) {
+            return LLVMNativePointer.create(LLVMStack.allocateUniqueStackMemory(frame, memory, this, getStackPointerSlot(), getSize(), getAlignment()));
         }
     }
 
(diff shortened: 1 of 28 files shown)


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/memory/LLVMAllocInstruction.java
```diff
@@ -88,13 +89,28 @@ public int getLength() {
...
+
         @Specialization
         protected LLVMNativePointer doOp(VirtualFrame frame,
                         @Cached("getLLVMMemory()") LLVMMemory memory) {
             return LLVMNativePointer.create(LLVMStack.allocateStackMemory(frame, memory, getStackPointerSlot(), getSize(), getAlignment()));
         }
     }
 
+    @NodeField(type = UniqueSlot.class, name = "uniqueSlot")
+    public abstract static class LLVMUniqueAllocConstInstruction extends LLVMAllocConstInstruction {
```
rschatz: As far as I understand your PR, this is not actually "allocating" anything, but returning the pointer to a pre-allocated stack location, right?
In that case, the node should probably not be called `...Alloc...Instruction`, but rather `LLVMGetUniqueStackPointerInstruction` or something like that.

anatol1234: > As far as I understand your PR, this is not actually "allocating" anything, but returning the pointer to a pre-allocated stack location, right?

Exactly.

> In that case, the node should probably not be called ...Alloc...Instruction, but rather LLVMGetUniqueStackPointerInstruction or something like that.

I agree that `LLVMUniqueAllocConstInstruction` does not fit well. The problem is, since the `LLVMUniqueAllocConstInstruction` also IS an `LLVMAllocInstruction`, one could argue that consequently the superclasses `LLVMAllocInstruction` and `LLVMAllocConstInstruction` need to be renamed for a similar reason. Those classes provide functionality for stack allocated memory (`getSize()`, `getAlignment()`, `getOffsets()` etc.) which is independent from whether it is applied to newly allocated memory or pre-allocated memory (that is why I made my node part of this type hierarchy). This ambivalence between "merely getting a pointer to memory" and "allocating memory" should be reflected by the name then. I thought about expressing this using words like "claim" (close to "allocate", but could also be understood in the sense of "claiming a right on something which already belongs to the one claiming" ) or "obtain" (close to "get", but expresses that there is potentially a bit of additional effort - allocation in our context - involved).

How about:
`LLVMAllocInstruction` -> `LLVMClaimStackSpaceInstruction`
`LLVMAllocConstInstruction` -> `LLVMClaimStackForConstInstruction`
`LLVMUniqueAllocConstInstruction` -> `LLVMGetUniqueStackPointerInstruction`

Things like `createUniqueAlloc` would also need to be renamed accordingly.

rschatz: I actually like "get" better than "claim" or "obtain" for the base class. It's simply getting a pointer to some stack memory. For some subclasses it's newly allocated memory, for some subclasses it's preallocated.

How about:
`LLVMAllocInstruction` -> `LLVMGetStackSpaceInstruction`
leave `LLVMAllocaInstruction` and `LLVMAllocaConstInstruction` (with the "a", they represent the `alloca` bitcode and `alloca` C function respectively)
`LLVMUniqueAllocConstInstruction` -> `LLVMGetUniqueStackSpaceInstruction`

anatol1234: Ok. I pushed changes implementing the naming scheme you suggested. The commit also includes the requested documentation work on `UniquesRegion` and the removal of code duplication (`setMSB` -> `Integer.highestOneBit`).

rschatz: Thanks, lgtm.

PR #855, opened 2018-07-04
Fix some small problems

pointhi:
* Add files to .gitignore which are build artifacts
* Fix small issue in README
* Force local on default, to fix some issues with local dependent testcases


pointhi:
@rschatz I think I made a nice solution for the environment problem

rschatz:
I like that solution. Running internal gate...

PR #859, opened 2018-07-09
Add On-Stack Replacement for Sulong.

rmosaner:
This PR adds On-Stack Replacement (OSR) for Sulong.

Based on the block based structure of LLVM IR bitcodes, loops are detected and for each loop a new node is created, which conforms to Truffle interface for loop nodes and replaces the basic block which was identified as the header of the loop.

Thus, Graal can replace loops during (interpreted) execution with optimized (compiled) versions, yielding a faster warmup. However, due to additional nodes, peak performance can be negatively affected.

Parameters (with default values) for playing around with OSR:
- -Dgraal.TruffleOSR=true ... toggle OSR
- -Dgraal.TruffleOSRCompilationThreshold=100000 ... threshold for loop iterations after which a loop is OSR compiled


projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMDispatchBasicBlockNode.java
```diff
@@ -93,11 +98,36 @@ public Object executeGeneric(VirtualFrame frame) {
...
         outer: while (basicBlockIndex != LLVMBasicBlockNode.RETURN_FROM_FUNCTION) {
             CompilerAsserts.partialEvaluationConstant(basicBlockIndex);
-            LLVMBasicBlockNode bb = bodyNodes[basicBlockIndex];
+
+            if(bodyNodes[basicBlockIndex] instanceof LLVMLoopNode) {
+                LLVMLoopNode loop = (LLVMLoopNode) bodyNodes[basicBlockIndex];
+                loop.execute(frame);
+
+                Integer[] successors = loop.getSuccessors();
+                for(int i = 0; i < successors.length; i++) {
```
christianhaeubl: the iteration should stop at successors.length - 1 and there should be some explicit code to handle the the last successor (see the pattern used for the LLVMSwitchNode below). Otherwise, the graph will look a bit worse.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
...
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMControlFlowNode;
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMExpressionNode;
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMStatementNode;
+
+public final class LLVMLoopDispatchNode extends LLVMExpressionNode {
+
+    private final FrameSlot exceptionValueSlot;
+    private final int headerId;
+    @Children private final LLVMStatementNode[] bodyNodes;
+    @CompilationFinal private final Integer[] indexMapping;
```
christianhaeubl: any reason why you are using Integer instead of int? (same applies to loopSuccessors).
With Integer[], you would need @CompilationFinal(dimensions = 1) - otherwise, the values are not compilation final (also applies to loopSuccessors).

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
...
+        this.indexMapping = indexMapping;
+        this.headerId = headerId;
+        this.loopSuccessors = successors;
+        this.successorSlot = successorSlot;
+    }
+
+    @ExplodeLoop
+    private boolean isInLoop(int bci) {
+        for(int i : loopSuccessors) {
+            if(i == bci)
```
christianhaeubl: please always use braces

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMBreakException.java
```diff
@@ -0,0 +1,45 @@
...
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.oracle.truffle.llvm.nodes.control;
+
+import com.oracle.truffle.api.nodes.ControlFlowException;
+
+public class LLVMBreakException extends ControlFlowException{
```
jkreindl: Please make this class `final`.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopNode.java
```diff
@@ -0,0 +1,88 @@
...
+import com.oracle.truffle.api.nodes.Node;
+import com.oracle.truffle.api.nodes.RepeatingNode;
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMExpressionNode;
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMStatementNode;
+
+public class LLVMLoopNode extends LLVMStatementNode {
+
+    @Child private LoopNode loop;
+
+    @CompilationFinal private final Integer[] successors;
```
jkreindl: I suspect there's a reason why this isn't an `int[]`. Please document it in a comment.

christianhaeubl: see other comment regarding int vs. Integer and @CompilationFinal(dimensions = 1)

jkreindl: Are the entries of this array constant as well? If so, you should change this to `@CompilationFinal(dimensions = 1)`.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopNode.java
```diff
@@ -0,0 +1,88 @@
...
+        public LLVMRepeatingNode(LLVMExpressionNode bodyNode) {
+            this.bodyNode = bodyNode;
+        }
+
+        @Override
+        public boolean executeRepeating(VirtualFrame frame) {
+            Object ret = null;
+
+            try {
+                ret = bodyNode.executeGeneric(frame);
```
christianhaeubl: just pull the return statement into the try-catch

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopNode.java
```diff
@@ -0,0 +1,88 @@
...
+public class LLVMLoopNode extends LLVMStatementNode {
+
+    @Child private LoopNode loop;
+
+    @CompilationFinal private final Integer[] successors;
+
+    public LLVMLoopNode(LLVMExpressionNode bodyNode, Integer[] successorIDs) {
+        loop = Truffle.getRuntime().createLoopNode(new LLVMRepeatingNode(bodyNode));
+
+        successors = Arrays.copyOf(successorIDs, successorIDs.length);
```
christianhaeubl: normally, there shouldn't be any need to copy arrays as the content of the array should not be modified later on anyways.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMDispatchBasicBlockNode.java
```diff
@@ -93,11 +98,36 @@ public Object executeGeneric(VirtualFrame frame) {
...
+                            continue outer;
+                        }
+                    } catch (FrameSlotTypeException e) {
+                        CompilerDirectives.transferToInterpreter();
+                        throw new RuntimeException("Error while reading from loop successor frame slot - type mismatch!");
+                    }
+                }
+
+                CompilerDirectives.transferToInterpreter();
+                throw new IllegalStateException("Must not reach here!");
```
jkreindl: Why not? You could make this error message more descriptive by e.g. stating that you did not find a loop successor.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -71,6 +77,7 @@
     private final Source source;
     private final LazyFunctionParser parser;
     private final DebugInfoFunctionProcessor diProcessor;
+    private FrameSlot loopSuccessorSlot;
```
christianhaeubl: do you really need this field? seems to be used only in one method

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMDispatchBasicBlockNode.java
```diff
@@ -93,11 +98,36 @@ public Object executeGeneric(VirtualFrame frame) {
...
+                Integer[] successors = loop.getSuccessors();
+                for(int i = 0; i < successors.length; i++) {
+                    try {
+                        if(frame.getInt(loopSuccessorSlot) == successors[i]) {
+                            basicBlockIndex = successors[i];
+                            continue outer;
+                        }
+                    } catch (FrameSlotTypeException e) {
+                        CompilerDirectives.transferToInterpreter();
+                        throw new RuntimeException("Error while reading from loop successor frame slot - type mismatch!");
```
jkreindl: You can include `e` as cause for the `RunTimeException` in the call to its constructor.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
                         method.getParameters().size(), symbols,
                         method, liveness, notNullable, dbgInfoHandler);
         method.accept(visitor);
+
+        LLVMControlFlowGraph cfg = new LLVMControlFlowGraph(method.getBlocks().toArray(new InstructionBlock[method.getBlocks().size()]));
+        try {
+            cfg.build();
+        } catch (Exception e) {
```
christianhaeubl: catching exception is a bit too generic - this will hide errors in the implementation of the loop detection - actually, I don't see any reason why the build method should throw an exception at all (it should just return and set the isReducible flag of cfg).

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
...
+                Integer[] successors = loop.getSuccessors();
+                for(int i = 0; i < successors.length; i++) {
+                    try {
+                        if(frame.getInt(successorSlot) == successors[i]) {
+                            basicBlockIndex = successors[i];
+                            continue outer;
+                        }
+                    } catch (FrameSlotTypeException e) {
+                        CompilerDirectives.transferToInterpreter();
+                        throw new RuntimeException("Error while reading from loop successor frame slot - type mismatch!");
```
jkreindl: I already mentioned this above, but a more descriptive error message would come in handy for debugging. In general, since this class is in large parts a duplicate of `LLVMDispatchBasicBlockNode` it would be preferable to have the common parts in a common baseclass, but it may be hard to accomplish this without any performance regressions. Perhaps you could look into this as future work.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
...
+
+            LLVMStatementNode loop = runtime.getNodeFactory().createLoop(loopBody, l.getSuccessorIDs());
+
+            // replace header block with loop node
+            nodes.remove(headerId);
+            nodes.add(headerId, loop);
+            nodeArray[headerId] = loop;
+        }
+
+        return runtime.getNodeFactory().createFunctionBlockNode(frame.findFrameSlot(LLVMUserException.FRAME_SLOT_ID), nodes, uniquesRegion.build(), nullableBeforeBlock, nullableAfterBlock,
```
christianhaeubl: with some minor refactoring, it should be possible to have the call to createFunctionBlockNode in the caller (shared with the one call that is already there)

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopNode.java
```diff
@@ -0,0 +1,88 @@
...
+
+        @Child private LLVMExpressionNode bodyNode;
+
+        public LLVMRepeatingNode(LLVMExpressionNode bodyNode) {
+            this.bodyNode = bodyNode;
+        }
+
+        @Override
+        public boolean executeRepeating(VirtualFrame frame) {
+            Object ret = null;
```
jkreindl: You could replace this method body with `return bodyNode.executeI1(frame)`. Instead of a `RuntimeException` that would also throw the IMHO more fitting `UnexpectedResultException`.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
                         method.getParameters().size(), symbols,
                         method, liveness, notNullable, dbgInfoHandler);
         method.accept(visitor);
+
+        LLVMControlFlowGraph cfg = new LLVMControlFlowGraph(method.getBlocks().toArray(new InstructionBlock[method.getBlocks().size()]));
```
jkreindl: It is not necessary to pass a preallocated array to `toArray`. Its safer to instead use a constant empty array like `FunctionDefinition.EMPTY`.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
...
         method.onAfterParse();
 
         return Truffle.getRuntime().createCallTarget(rootNode);
     }
 
+    private LLVMExpressionNode resolveLoops(List<LLVMStatementNode> nodes, LLVMControlFlowGraph cfg, FrameDescriptor frame, FrameSlot[][] nullableBeforeBlock,
+                    FrameSlot[][] nullableAfterBlock, LLVMSourceLocation location, LLVMStatementNode[] copyArgumentsToFrameArray, UniquesRegion uniquesRegion) {
+
+        if (cfg.getCFGLoops().size() > 0) {
+            loopSuccessorSlot = frame.addFrameSlot("__LLVMLoopSuccessor", FrameSlotKind.Int);
```
jkreindl: Sulong-internal frameslots usually use an identifier in the form "<name>". See e.g. `LLVMStack.FRAME_ID` or `LLVMUserException.FRAME_SLOT_ID`. This is by no means an explicitly enforced convention, but I think doing this avoids name conflicts and makes debugging easier.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
...
+        }
+
+        LLVMStatementNode[] nodeArray = nodes.toArray(new LLVMStatementNode[0]);
```
jkreindl: Use `LLVMStatementNode.NO_STATEMENTS` instead of explicitly allocating an array here.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
...
+            if(i == bci)
+                return false;
+        }
+        return true;
+    }
+
+    @Override
+    @ExplodeLoop(kind = LoopExplosionKind.MERGE_EXPLODE)
+    public Object executeGeneric(VirtualFrame frame) {
+        boolean ret = false; // helper variable to not return at first encounter of headerId
```
christianhaeubl: something like firstIteration would be more readable and would make the comment unnecessary

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
...
+
+        for (CFGLoop l : cfg.getCFGLoops()) {
```
jkreindl: Please avoid single-letter variable names if the values are used more often or not obviously in your loop body.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
...
+
+        CompilerAsserts.compilationConstant(bodyNodes.length);
+        int basicBlockIndex = headerId;
+
+        outer: while (isInLoop(basicBlockIndex) && (basicBlockIndex != headerId || !ret)) { // do-while loop fails at PE
```
christianhaeubl: in general, comments are more readable if they are on a separate line

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
...
+                indexMapping[b.id] = i++;
+            }
+
+            LLVMExpressionNode loopBody = runtime.getNodeFactory().createLoopDispatchNode(frame.findFrameSlot(LLVMUserException.FRAME_SLOT_ID), Collections.unmodifiableList(bodyNodes),
+                            nullableBeforeBlock, nullableAfterBlock, headerId, indexMapping, l.getSuccessorIDs(), loopSuccessorSlot);
+
+            LLVMStatementNode loop = runtime.getNodeFactory().createLoop(loopBody, l.getSuccessorIDs());
+
+            // replace header block with loop node
+            nodes.remove(headerId);
```
jkreindl: You could make removing and adding an atomic operation by calling `nodes.set(headerId, loop)` instead.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
...
+            CompilerAsserts.partialEvaluationConstant(basicBlockIndex);
+
+            ret = true;
+
+            if(bodyNodes[indexMapping[basicBlockIndex]] instanceof LLVMLoopNode) {
```
christianhaeubl: see comment in LLVMDispatchBasicBlockNode

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
...
+            List<LLVMStatementNode> bodyNodes = new ArrayList<>();
+
+            // add header to body nodes
+            bodyNodes.add(nodes.get(headerId));
+            indexMapping[headerId] = 0;
+
+            // add body nodes
+            int i = 1;
+            for (CFGBlock b : l.getBody()) {
+                bodyNodes.add(nodeArray[b.id]);
```
jkreindl: Since you keep `nodeArray` and `nodes` in sync in any case, I don't see a need for `nodeArray`.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
...
+        if(basicBlockIndex != headerId) {
+            frame.setInt(successorSlot, basicBlockIndex);
+            return false;
+        }
+
+        return true;
+    }
+
+    @ExplodeLoop
+    private static void executePhis(VirtualFrame frame, LLVMControlFlowNode controlFlowNode, int successorIndex) {
```
christianhaeubl: share the helper functions that are down here with the LLVMDispatchBasicBlockNode (this will not impact PE)

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/util/LLVMControlFlowGraph.java
```diff
@@ -0,0 +1,335 @@
...
+            return;
+
+        active.add(loop);
+        for (CFGBlock b : loop.body) {
+            if (b.isLoopHeader) {
+                CFGLoop inner = cfgLoops[b.loopId];
+                if (active.contains(inner)) {
+                    // catches case that there is a stack overflow because two loop nodes are being called iteratively
+                    // from one another, without one being left beforehand
+                    throw new RuntimeException("Irreducible nestedness!");
```
christianhaeubl: throw a custom exception and catch it in the build method (applies to all bailouts in here)

christianhaeubl (changes requested):

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/util/LLVMControlFlowGraph.java
```diff
@@ -0,0 +1,335 @@
...
+import com.oracle.truffle.llvm.parser.model.blocks.InstructionBlock;
+import com.oracle.truffle.llvm.parser.model.symbols.instructions.TerminatingInstruction;
+
+public final class LLVMControlFlowGraph {
+
+    private static final int LOOP_CONTAINER_MAX_CAPACITY = Long.SIZE;
+    private static final int LOOP_HEADER_INITIAL_CAPACITY = 4;
+
+    private CFGBlock[] blocks;
+    private CFGLoop[] cfgLoops = new CFGLoop[LOOP_HEADER_INITIAL_CAPACITY];
```
jkreindl: Do you really need an array that you resize manually? I think you should use an `ArrayList` instead, if not to make your code simpler than at least to make it safer.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
```diff
@@ -109,21 +117,83 @@ public RootCallTarget convert() {
...
+                indexMapping[b.id] = i++;
+            }
+
+            LLVMExpressionNode loopBody = runtime.getNodeFactory().createLoopDispatchNode(frame.findFrameSlot(LLVMUserException.FRAME_SLOT_ID), Collections.unmodifiableList(bodyNodes),
```
jkreindl: `getSuccessorIDs` allocates a new array with the same content on each call. Since, I assume, the contents of the returned arrays are expected to be identical, you should only call the method once and pass the same array to `createLoop` and `createLoopDispatchNode`.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/util/LLVMControlFlowGraph.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+
+        }
+
+        public Integer[] getSuccessorIDs() {
+            if (successors == null) {
+                calculateSuccessors();
+            }
+
+            Integer[] sucIDs = new Integer[successors.size()];
```
jkreindl: Is the order of elements in `sucIDs` guaranteed by your loop detection algorithm? These seem like they should be sorted in some way.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/util/LLVMControlFlowGraph.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+        }
+
+    }
+
+    public List<CFGLoop> getCFGLoops() {
+        List<CFGLoop> loops = new ArrayList<>();
+
+        for (CFGLoop l : cfgLoops) {
+            if (l != null)
```
jkreindl: The style gate will probably catch this. Please always use braces to delimit the bodies of conditional branches even if they consist only of a single line.

projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/util/LLVMControlFlowGraph.java
```diff
@@ -0,0 +1,335 @@
...
+                loops.add(l);
+        }
+        loops.removeIf(l -> l == null);
```
jkreindl: This statement is unnecessary since you already make sure not to add any `null` elements to the list.

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMBreakException.java
christianhaeubl: is this class unused?

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMLoopDispatchNode.java
```diff
@@ -0,0 +1,265 @@
+/*
+ * Copyright (c) 2017, 2018, Oracle and/or its affiliates.
```
christianhaeubl: make sure that you have the correct copyright year in all files that you added

projects/com.oracle.truffle.llvm.nodes/src/com/oracle/truffle/llvm/nodes/control/LLVMDispatchBasicBlockNode.java
```diff
@@ -99,36 +100,26 @@ public Object executeGeneric(VirtualFrame frame) {
...
-                LLVMLoopNode loop = (LLVMLoopNode) bodyNodes[basicBlockIndex];
-                loop.execute(frame);
-                int successorBasicBlockIndex = FrameUtil.getIntSafe(frame, loopSuccessorSlot);
+            LLVMControlFlowNode controlFlowNode;
+            LLVMBasicBlockNode bb = null;
+            /*
+             * Distinction between low level basic blocks and high level control flow
+             * structures, like LLVMLoopNode
+             */
+            if(bodyNodes[basicBlockIndex] instanceof LLVMBasicBlockNode) {
```
jkreindl: This check will probably have a noticeable impact on performance. It should not be necessary either.

jkreindl (changes requested):

PR #863, opened 2018-07-24
Add Support for Protected Main Functions

sudosalim:
Compilers like the Swift one generates main in the form of
`define protected i32 @main`
which is currently not identified as main entry by Sulong but is visible by LLVM lli.

### DIFF
--- a/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/model/functions/FunctionDefinition.java
+++ b/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/model/functions/FunctionDefinition.java
@@ -265,7 +265,9 @@ public final class FunctionDefinition implements Constant, FunctionSymbol, Metad
 
     @Override
     public boolean isExported() {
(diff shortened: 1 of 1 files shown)


lukasstadler:
Thanks for the PR!

@christianhaeubl / @jkreindl - maybe this can also be solved at the point where we look up the main method (as opposed to the export logic, like in this PR), so that not only exported functions are considered?

jkreindl:
According to [the LLVM Language Reference](https://llvm.org/docs/LangRef.html#visibility-styles) the semantics of visibility styles such as `protected` are platform-dependant. We currently do not consider that at all. For the ELF case any `protected` function should be exported, only it must not be overwritten for any function within the defining module. I would not merge this PR because we should really fix our dynamic linker instead of just introducing a special case here. @christianhaeubl what is your opinion on this.

christianhaeubl:
Thanks for reporting the issue - Sulong did not handle the `protected` visibility correctly when doing the linking. I fixed the issue in ce43ebfd7b76940d1057d51a803f5cb235973e34

